Issue #4, opened 2019-09-18
Http client for validator REST api


mpetrunic:
- create client for this api: https://github.com/ethereum/eth2.0-APIs/tree/master/apis/validator
- use axios to make request
- use axios suggested error handling and convert to custom exception class

PR #11, opened 2019-09-27
Feature/client rest

jogamod:
Implemented Beacon API Client for Validator REST API.  
resolves Issue #4.


package.json
```diff
@@ -113,6 +114,8 @@
     "@types/react-test-renderer": "^16.8.1",
     "@types/webdriverio": "^4.8.7",
     "@types/webpack-env": "^1.13.3",
+    "axios": "0.19.0",
```
mpetrunic: should not be devDependency

src/renderer/constants/apiUrls.ts
```diff
@@ -0,0 +1,41 @@
+import { Epoch, Slot, Shard, BLSPubkey, BeaconBlock, uint8, IndexedAttestation } from '@chainsafe/eth2.0-types';
+
+export const API_URL: String = `https://${process.env.HOSTNAME || 'localhost'}:${process.env.PORT || '3000'}`;
```
mpetrunic: create config file which will contain all configuration and default values if env is missing.
Also API_URL should be passed as param to api client as we will instantiate multiple api client with different url at the same time

src/renderer/constants/apiUrls.ts
```diff
@@ -0,0 +1,41 @@
...
+ * a validator to connect and perform its obligations on the Ethereum
+ * 2.0 phase 0 beacon chain.
+ */
+
+export const FETCH_NODE_VERSION: string = `${API_URL}/node/version`;
+export const FETCH_GENESIS_TIME: string = `${API_URL}/node/genesis_time`;
+export const POLL_NODE_SYNCING: string = `${API_URL}/node/syncing`;
+export const FETCH_FORK_INFORMATION: string = `${API_URL}/node/fork`;
+
+export const FETCH_VALIDATOR_DUTIES: Function = (validator_pubkeys: BLSPubkey[], epoch: Epoch) => {
```
mpetrunic: keep variables as camelCase. I will add to eslint if there isn't already

src/renderer/constants/apiUrls.ts
```diff
@@ -0,0 +1,41 @@
...
+    return `${API_URL}/validator/duties?validator_pubkeys=${validator_pubkeys}&epoch=${epoch}`;
```
mpetrunic: I'm pretty sure this doesn't work as pubkeys are buffers and should be serialized as array of hex strings.
Something like this might work:
```javascript
const hexPubKeys = validatorPublicKeys.map(key => key.toString("hex"));
const url = `/duties?validator_pubkeys=${JSON.stringify(hexPubKeys)}&epoch=${epoch.toString()}`;
```

src/renderer/constants/apiUrls.ts
```diff
@@ -0,0 +1,41 @@
...
+};
+
+export const FETCH_VALIDATOR_BLOCK: Function = (slot: Slot, randao_reveal: string) => {
+    return `${API_URL}/validator/block?slot=${slot}&randao_reveal=${randao_reveal}`;
+};
+
+export const PUBLISH_SIGNED_BLOCK: Function = (beacon_block: BeaconBlock) => {
+    return `${API_URL}/validator/block?beacon_block=${beacon_block}`;
```
mpetrunic: beacon block should be in body as this is POST request

src/renderer/constants/apiUrls.ts
```diff
@@ -0,0 +1,41 @@
...
+    validator_pubkey: BLSPubkey,
+    poc_bit: uint8,
+    slot: Slot,
+    shard: Shard
+) => {
+    return `${API_URL}/validator/attestation?validator_pubkey=${validator_pubkey}&poc_bit=${poc_bit}&slot=${slot}&shard=${shard}`;
+};
+
+export const PUBLISH_SIGNED_ATTESTATION: Function = (attestation: IndexedAttestation) => {
+    return `${API_URL}/validator/attestation?attestation=${attestation}`;
```
mpetrunic: should also be in body as it's POSt request

src/renderer/constants/apiUrls.ts
```diff
@@ -0,0 +1,41 @@
...
+    return `${API_URL}/validator/block?beacon_block=${beacon_block}`;
+};
+
+export const PRODUCE_ATTESTATION: Function = (
+    validator_pubkey: BLSPubkey,
+    poc_bit: uint8,
+    slot: Slot,
+    shard: Shard
+) => {
+    return `${API_URL}/validator/attestation?validator_pubkey=${validator_pubkey}&poc_bit=${poc_bit}&slot=${slot}&shard=${shard}`;
```
mpetrunic: pubkey should be converted to hex

src/renderer/services/BeaconAPIClient.ts
```diff
@@ -0,0 +1,226 @@
...
+    fetchValidatorBlock(slot: Slot, randao_reveal: string): Promise<BeaconBlock>;
+    publishSignedBlock(beacon_block: BeaconBlock): Promise<any>;
+    produceAttestation(
+        validator_pubkey: BLSPubkey,
+        poc_bit: uint8,
+        slot: Slot,
+        shard: Shard
+    ): Promise<IndexedAttestation>;
+    publishSignedAttestation(attestation: IndexedAttestation): Promise<any>;
+}
```
mpetrunic: extract interfaces to separate file called interface.ts

src/renderer/services/BeaconAPIClient.ts
```diff
@@ -0,0 +1,226 @@
...
+
+export class BeaconAPIClient implements BeaconAPIClientInterface {
```
mpetrunic: needs to have constructor like

```
interface IBeaconApiClientOptions {
   url: string;
}

public constructor(options: IBeaconApiClientOptions)
```

src/renderer/services/BeaconAPIClient.ts
```diff
@@ -0,0 +1,226 @@
...
+        return result;
+    }
+
+    /**
+     * Helper function for POST method
+     * @param url endpoint url
+     * @param data request body
+     * @param callback callback function that will return either response or error
+     */
+    postToApi = (url: string, data: any = {}, callback: Function) => {
```
mpetrunic: It might be better to extract those generic http api methods to separate class like here:
https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar/src/util/httpClient.ts

package.json
jogamod: all change requests done

src/renderer/constants/apiUrls.ts
```diff
@@ -0,0 +1,34 @@
+import { Epoch, Slot, Shard, BLSPubkey, BeaconBlock, uint8, IndexedAttestation } from '@chainsafe/eth2.0-types';
+import { DEFAULT_HOSTNAME, DEFAULT_PORT } from './defaultConstants';
+
+export const API_URL: string = `https://${process.env.HOSTNAME || DEFAULT_HOSTNAME}:${process.env.PORT ||
```
mpetrunic: I think you don't need this

src/renderer/services/httpClient.ts
```diff
@@ -0,0 +1,51 @@
+import axios, { AxiosError, AxiosInstance, AxiosResponse } from 'axios';
+
+export class HttpClient {
+    private client: AxiosInstance;
+
+    public constructor(opt: Partial<string>) {
+        this.client = Axios.create({
+            baseURL: opt || ''
```
mpetrunic: I think you should throw error if url is missing

src/renderer/components/Counter.tsx
```diff
@@ -15,7 +15,7 @@ const Counter: React.FunctionComponent<Props> = ({ value, incrementValue, decrem
         <p>
             <img src={redCubeImg} />
         </p>
-        <p id="counter-value">Current value: {value}</p>
+        <p id="counter-value">Current value: {value} </p>
```
mpetrunic: this space is failing frontend test

src/renderer/services/BeaconAPIClient.ts
```diff
@@ -0,0 +1,136 @@
...
+        this.httpClient = new HttpClient(options.urlPrefix);
+    }
+
+    async fetchNodeVersion(): Promise<string> {
+        // TODO handle error catch
+        try {
+            const result = await this.httpClient.get<string>(FETCH_NODE_VERSION);
+            return result;
+        } catch (error) {
+            return error;
```
mpetrunic: Just noticed this. I would remove error and let whoever is using library to catch them. You are returning error so you would have to change return type of method to be `Promise<string|Error>`.

jogamod: Will refactor this in an hour.

src/renderer/services/BeaconAPIClient.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    constructor(options: IBeaconApiClientOptions) {
+        if (!options.urlPrefix) {
+            throw new EmptyUrlError();
+        }
+        this.options = options;
+        this.httpClient = new HttpClient(options.urlPrefix);
+    }
+
+    async fetchNodeVersion(): Promise<string | Error> {
```
mpetrunic: damm, commit before was ok, you won't actually return Error it will get thrown just leave string. You can inline result variable as well

jogamod: True that.

PR #12, opened 2019-09-27
LevelDB Wrapper

aostrun:
Abstract DB wrapper with LevelDB implementation.
Tests included.


package.json
```diff
@@ -103,19 +103,31 @@
     "@babel/preset-env": "^7.4.5",
     "@babel/preset-react": "^7.0.0",
     "@babel/preset-typescript": "^7.3.3",
+    "@chainsafe/eth2.0-config": "^0.1.0",
```
mpetrunic: should not be devDependency

package.json
```diff
@@ -103,19 +103,31 @@
...
+    "@chainsafe/eth2.0-types": "^0.1.0",
+    "@chainsafe/ssz": "^0.5.1",
```
mpetrunic: should not be devDependency

package.json
```diff
@@ -103,19 +103,31 @@
...
     "@types/react-test-renderer": "^16.8.1",
+    "@types/sinon": "^7.0.13",
     "@types/webdriverio": "^4.8.7",
     "@types/webpack-env": "^1.13.3",
     "babel-loader": "^8.0.6",
+    "chai": "^4.2.0",
+    "chai-as-promised": "^7.1.1",
     "cross-env": "^5.1.3",
     "css-loader": "^2.1.1",
+    "deepmerge": "^4.0.0",
```
mpetrunic: should not be devDependency

src/db/api/abstract.ts
```diff
@@ -0,0 +1,26 @@
...
+import { IService } from '../../interfaces';
+
+export interface IDatabaseApiOptions {
+    config: IBeaconConfig;
+    controller: IDatabaseController;
+}
+
+export abstract class DatabaseService implements IService {
+    protected config: IBeaconConfig;
+    protected db: IDatabaseController;
```
mpetrunic: both properties could have readonly modifier

src/db/controller/impl/level.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    private opts: ILevelDBOptions;
+
+    private logger: ILogger;
+
+    public constructor(opts: ILevelDBOptions, { logger }: { logger: ILogger }) {
+        super();
+        this.opts = opts;
+        this.logger = logger;
+        this.db = opts.db || level(opts.name || 'beaconchain', { keyEncoding: 'binary', valueEncoding: 'binary' });
```
mpetrunic: don't think beaconchain is correct default name :)

src/db/options.ts
```diff
@@ -0,0 +1,27 @@
+import { IConfigurationModule } from '../util/config';
+
+export interface IDatabaseOptions {
+    name: string;
+}
+
+export const DatabaseOptions: IConfigurationModule = {
```
mpetrunic: We don't need this, this is for cli

src/db/schema.ts
```diff
@@ -0,0 +1,58 @@
...
+    validator,
+    deposit, // index -> Deposit
+    exit, // hash -> VoluntaryExit
+    transfer, // hash -> Transfer
+    proposerSlashing, // hash -> ProposerSlashing
+    attesterSlashing, // hash -> AttesterSlashing
+    merkleTree, // depositIndex -> MerkleTree
+    // validator
+    lastProposedBlock,
+    proposedAttestations
```
mpetrunic: clear those buckets

src/db/schema.ts
```diff
@@ -0,0 +1,58 @@
...
+export enum Key {
+    chainHeight,
+
+    latestState,
+    finalizedState,
+    justifiedState,
+
+    finalizedBlock,
+    justifiedBlock,
+    progressiveMerkleTree
```
mpetrunic: clear those keys

src/logger/winston.ts
```diff
@@ -0,0 +1,118 @@
+/**
+ * @module logger
+ */
+
+import { createLogger, format, Logger, transports } from 'winston';
+import { defaultLogLevel, LogLevel, ILogger, ILoggerOptions } from './interface';
+import chalk from 'chalk';
+
+export class WinstonLogger implements ILogger {
```
mpetrunic: We don't wan't to use winston as we need something cross OS compatible (storing location and managing rotation on different OS) so don't implement logger interface just yet

src/util/config.ts
```diff
@@ -0,0 +1,78 @@
+export interface IConfigurationField<T = unknown> {
```
mpetrunic: we don't need this file entirely, it's for CLI

src/db/api/repository.ts
```diff
@@ -0,0 +1,78 @@
...
+    public async has(id: Id): Promise<boolean> {
+        return (await this.get(id)) !== null;
+    }
+
+    public async setUnderRoot(value: T): Promise<void> {
+        await this.set(hashTreeRoot(value, this.type), value);
+    }
+
+    public async set(id: Id, value: T): Promise<void> {
+        await this.db.put(encodeKey(this.bucket, id), serialize(value, this.type));
```
mpetrunic: could you extract serialize/deserialize/hashTreeRoot into separate class (something like "CGSerialization) so we can change serialization to google protobufs at some point.

mpetrunic:
@aostrun I think all those things should be under /src/renderer/services

aostrun:
Fixed all issues and implemented `CGSerialization<T>` interface. I tried creating a generic interface for serialization where type T represents a wrapper for all types supported by the serializer implementation. E.g. `SSZSerializer` has type T set to `AnySSZType`.
I'm not too happy with how this ends up looking inside the abstract `Repository` class as we have to set the serializer dependency `serializer: CGSerialization<any>`  generic type to `any` as we want to support various serializers.
@mpetrunic What do you think about this?

src/renderer/services/db/abstract.ts
```diff
@@ -0,0 +1,7 @@
+import { object } from 'prop-types';
+
+export interface CGSerialization<T> {
+    serialize(value: any, type: T): Buffer;
+    deserialize(value: Buffer, type: T): any;
```
mpetrunic: I will answer you question here. We haven't really solved this one in ssz library as generics in typescript are kinda limited. So we can leave `any` here. 

I would only suggest for deserialize method to be
`deserialize<R>(value: Buffer, type: T): R`
as it's much easier to set type than doing `as Something` everywhere

mpetrunic: Also you can hardcode AnySSZType as type in CGSerializer (you don't need generics) because if we change serialization we will write adapter to convert AnySSZType to required schema rather than writing type descriptors for every type

src/renderer/services/db/controller/impl/level.ts
```diff
@@ -0,0 +1,89 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/**
+ * @module db/controller/impl
+ */
+
+import { LevelUp } from 'levelup';
+import { IDatabaseController, ISearchOptions } from '../interface';
+import { EventEmitter } from 'events';
+// @ts-ignore
```
mpetrunic: you still need ignore even after you added level types?

src/renderer/services/db/schema.ts
```diff
@@ -0,0 +1,50 @@
...
+    block, // hash -> BeaconBlock
+    invalidBlock, // bad block
+    mainChain, // slot -> blockHash
+    chainInfo, // Key -> number64 | stateHash | blockHash
+    validator,
+    deposit, // index -> Deposit
+    exit, // hash -> VoluntaryExit
+    transfer, // hash -> Transfer
+    proposerSlashing, // hash -> ProposerSlashing
+    attesterSlashing // hash -> AttesterSlashing
```
mpetrunic: We can remove all Bucket values and keep it empty for now, until we start storing stuff

src/renderer/services/db/schema.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+export enum Key {
+    chainHeight,
+
+    latestState,
+    finalizedState,
+    justifiedState
```
mpetrunic: We can remove all Key values and keep it empty for now, until we start storing stuff

src/renderer/services/db/serializers/ssz.ts
```diff
@@ -0,0 +1,18 @@
...
+        return deserialize(value, type);
+    }
+    hashTreeRoot(value: any, type: AnySSZType): Buffer {
+        return hashTreeRoot(value, type);
+    }
+}
+
+const SSZ = new SSZSerializer();
+
+export default SSZ;
```
mpetrunic: don't use export default. I will try to update eslint ro prefer named export

Reasoning:
https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/

src/renderer/services/db/controller/impl/level.ts
aostrun: There are no types for `level` package, or at least I wasn't able to find them :/

mpetrunic: its ok then. I will create PR for fix of e2e tests with esModuleInterop flag. It will include a lot of style guide changes, hopefully it won't break things much

src/renderer/services/db/schema.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * @module db/schema
+ */
+
+// tslint:disable-next-line: import-name
+import BN from 'bn.js';
+
+// Buckets are separate database namespaces
+export enum Bucket {
+    test
```
mpetrunic: nope, hardcode some string manually in test

package.json
```diff
@@ -134,6 +131,7 @@
     "file-loader": "^3.0.1",
     "fork-ts-checker-webpack-plugin": "^1.3.4",
     "html-webpack-plugin": "^3.2.0",
+    "husky": "^2.3.0",
```
mpetrunic: we removed that

PR #13, opened 2019-09-29
Switch to eslint and improve CI

mpetrunic:
- switch to eslint
- imrpove CI
- enable esModuleInterop flag


src/renderer/store/index.ts
```diff
@@ -1,19 +1,19 @@
...
+import {RootState, rootReducer} from "../reducers";
 
 const configureStore = (initialState?: RootState): Store<RootState | undefined> => {
-    const middlewares: any[] = [];
-    const enhancer = composeWithDevTools(applyMiddleware(...middlewares));
+    const middlewares: Middleware[] = [],
+     enhancer = composeWithDevTools(applyMiddleware(...middlewares));
     return createStore(rootReducer, initialState, enhancer);
-};
+},
```
jogamod: Is this comma accidentally replaced with semicolon?

src/renderer/store/index.ts
```diff
@@ -1,19 +1,19 @@
...
 
-const store = configureStore();
+ store = configureStore();
```
jogamod: Where is the store declared?

.eslintrc.js
```diff
@@ -0,0 +1,55 @@
...
+        "@typescript-eslint"
+    ],
+    extends: [
+        "eslint:recommended",
+        "plugin:@typescript-eslint/recommended",
+        "plugin:react/recommended"
+    ],
+    "rules": {
+        "@typescript-eslint/class-name-casing": "error",
+        "@typescript-eslint/indent": "off",
```
jogamod: Maybe we should use indentation 2? It is very readable, but this is just my opinion.

PR #24, opened 2019-10-03
StoryBook init

### LINKED ISSUE #16: Setup storybook
https://storybook.js.org/docs/guides/guide-react/

Watch out! We are using typescript!
https://medium.com/@dandobusiness/setting-up-a-react-typescript-storybook-project-5e4e9f540568

morrigan:
- init StoryBook with TypeScript
 - load our sass style with StoryBook
 - created Button component with one type of style
 - created simple Button story for test

resolves #16


package.json
```diff
@@ -29,7 +29,7 @@
   "scripts": {
     "build-main": "cross-env NODE_ENV=production webpack --config webpack.main.prod.config.js",
     "build-renderer": "cross-env NODE_ENV=production webpack --config webpack.renderer.prod.config.js",
-    "build": "npm run build-main && npm run build-renderer",
+    "build": "npm run build-main && npm run build-renderer && build-storybook",
```
mpetrunic: I think it would be better to put build-storybook in "prestorybook" task instead of prolonging build time here

src/renderer/style/buttons.scss
```diff
@@ -0,0 +1,28 @@
+.button {
```
mpetrunic: should we put this in /components/buttons?

stories/Button.stories.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+import { storiesOf } from "@storybook/react";
+import { Button } from "../src/renderer/components/Button";
+
+storiesOf("Button", module)
```
mpetrunic: I wonder if it's better to put stories next to actual component?

package.json
morrigan: what would `prestorybook` be then? Tbh, I saw this from other repos. If this is used in CI, then better to remove.

mpetrunic: it's gonna execute everytime you call `yarn run storybook`

src/renderer/style/buttons.scss
morrigan: Yes, seems better and will have more buttons anyway.

stories/Button.stories.tsx
morrigan: That's possible too but I don't know if that's better. When I think of CSS, if everything won't be imported by `index.scss`  (but sometimes in the component specific for that file), then it is better.

Location-wise I don't have any feelings. If yes for this, then what about redux functions and tests?

mpetrunic: lets leave it like that. Not sure I understand thing about css?

morrigan: Nevermind the css but another thing occurred to me. We'll have to keep the same structure in `stories` directory as well as `src` regarding components/containers. I now kinda lean towards the suggested approach xD

mpetrunic: haha yeah, make sure that build doesnt include .stories.ts in output directory

package.json
morrigan: That build is not required for development. I don't think we even need build-storybook (will we display somewhere those static files)?

mpetrunic: Nah, but default build will compile those files because they are in src

morrigan:
@mpetrunic could you check the changes made? The CI tests fails for OS X.

mpetrunic:
> @mpetrunic could you check the changes made? The CI tests fails for OS X.

Let's merge my PR first and check if it fixes it?

package.json
```diff
@@ -41,7 +41,9 @@
     "test:e2e": "jest '(\\/test\\/e2e/).*'",
     "pack": "npm run build && electron-builder --dir",
     "dist": "npm run build && electron-builder",
-    "postinstall": "electron-builder install-app-deps"
+    "postinstall": "electron-builder install-app-deps",
+    "storybook": "start-storybook -p 6006",
+    "build-storybook": "build-storybook"
```
mpetrunic: you missed prestorybook task to build story book
`"prestorybook": "yarn run build-storybook"`

For the sake of consistency you might wanna rename build-storybook to build:storybook

morrigan:
@mpetrunic approval? :)

PR #25, opened 2019-10-10
Feature/validator signing key

jogamod:
Implemented wallet service for generating keypairs, signing, verifying and other functions that will be used in eth2.0 such as aggregate signatures, verify multiple signatures using `chainsafe/bls` library. 
Implemented Keystore like in eth1.0 style (have to be reimplemented when PR for eth2.0 is resolved). 
The address is generated using bech32 like @mpetrunic pointed out.


src/renderer/services/wallets/WalletInterfaces.ts
```diff
@@ -0,0 +1,29 @@
...
+import { PublicKey } from '@chainsafe/bls/lib/publicKey';
+import { Keypair } from '@chainsafe/bls/lib/keypair';
+import { Hash } from '@chainsafe/eth2.0-types';
+
+export type IKeystoreJson = V3Keystore;
+export type KeystoreParams = V3Params;
+
+export type Domain = Buffer;
+
+export interface IKeystore {
```
mpetrunic: I would suggest interface like this:
```javascript
export interface ICGKeystore {
   new(file: string);
   decrypt(password: string): Keypair;
   changePassword(oldPassword: string, newPassword: string): void;
   //deletes physical keystore file
   destroy(): void;
   static create(file: string, password: string, keypair: Keypair): Keystore;
}
```
@aostrun you can assume same keystore interface

you also need Eth2HDWallet class like:
```javascript
export interface IEth2HDWallet {
   // should return mnemonic
   static generate(entropy?: bytes): string;

   static getKeypair(mnemonic: string, walletIndex?: number = 0): Keypair;
}
```

src/renderer/services/wallets/eth1/Eth1Keystore.ts
```diff
@@ -0,0 +1,139 @@
...
+                    Buffer.from(password),
+                    kdfParams.salt,
+                    kdfParams.n,
+                    kdfParams.r,
+                    kdfParams.p,
+                    kdfParams.dklen
+                );
+                break;
+            default:
+                throw new Error('Unsupported kdf');
```
mpetrunic: Why did you copy all ethereumjs code.
You should be able to do something like:
```javascript
import {fromPrivateKey} from "ethereumjs-wallet";

const keystore = fromPrivateKey(keypair.privateKey.toBytesCompressed()).toV3();
return {
   ...keystore
   address: bech32.toWords(keypair.publicKey.toBytesCompressed())
}
```

src/renderer/services/Eth1ICGKeystore.ts
```diff
@@ -0,0 +1,101 @@
+import { ICGKeystore } from './interface';
+import { Keypair } from '@chainsafe/bls/lib/keypair';
+import * as fs from 'fs';
```
mpetrunic: import fs from 'fs' 
should work

src/renderer/services/Eth1ICGKeystore.ts
```diff
@@ -0,0 +1,101 @@
...
+import { PrivateKey } from '@chainsafe/bls/lib/privateKey';
+import { PublicKey } from '@chainsafe/bls/lib/publicKey';
+const eth1WalletProvider = require('ethereumjs-wallet');
+const bech32 = require('bech32');
```
mpetrunic: don't use require import.

src/renderer/services/Eth1ICGKeystore.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export class Eth1ICGKeystore implements ICGKeystore {
+    keystore: any;
```
mpetrunic: Type might be defined in ethereumjs-wallet. If not define props we need

src/renderer/services/Eth1ICGKeystore.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * Method used to decrypt encrypted private key from keystore file
+     * @param password
+     */
+    decrypt(password: string): Keypair {
+        const privateKeyString = eth1WalletProvider.fromV3(this.keystore, password).getPrivateKeyString();
+        const priv = PrivateKey.fromHexString(privateKeyString);
+        const pub = PublicKey.fromBytes(PublicKey.fromPrivateKey(priv).toBytesCompressed());
+        return new Keypair(priv, pub);
```
mpetrunic: just pass priv key. It's probably faster

src/renderer/services/Eth2HDWallet.ts
```diff
@@ -0,0 +1,28 @@
+import * as bip39 from 'bip39';
```
mpetrunic: import bip39 from 'bip39' ?

src/renderer/services/Eth2HDWallet.ts
```diff
@@ -0,0 +1,28 @@
...
+import { bytes } from '@chainsafe/eth2.0-types';
+import { PrivateKey } from '@chainsafe/bls/lib/privateKey';
+import { Keypair } from '@chainsafe/bls/lib/keypair';
+const hdkey = require('ethereumjs-wallet/hdkey');
```
mpetrunic: don't use require

test/components/HDWallet.spec.tsx
```diff
@@ -0,0 +1,33 @@
+import { Eth2HDWallet } from '../../src/renderer/services/Eth2HDWallet';
```
mpetrunic: this file should be in unit directory

test/components/Keystore.spec.tsx
```diff
@@ -0,0 +1,58 @@
...
+const privateKey = '0e43429c844ccedd4aff7aaa05fe996f41f9464b360ca03a4349387ba49b3e18';
+const privateKeyStr = `0x${privateKey}`;
+
+const keyStoreFilePath = `test/components/${getV3Filename()}.json`;
+const password = 'test';
+const newPassword = 'newTest';
+
+function getV3Filename(timestamp?: number) {
+    const ts = timestamp ? new Date(timestamp) : new Date();
+    return ['UTC--', ts.toJSON().replace(/:/g, '-'), '--', 'uuid'].join('');
```
mpetrunic: you should mock fs methods. Something like
const writeStub = sandbox.stub(fs, "writeFileSync")

src/renderer/services/Eth1ICGKeystore.ts
jogamod: Module '"fs"' has no default export. Issue https://github.com/Microsoft/TypeScript/issues/3337

src/renderer/services/Eth1ICGKeystore.ts
jogamod: There is not `@types/ethereumjs-wallet`, but there are `@types/bech32` so it is changed

src/renderer/services/Eth1ICGKeystore.ts
jogamod: Getting error `Module '"fs"' has no default export.`

src/renderer/services/Eth2HDWallet.ts
jogamod: Module '"/home/domagoj/NodeFactory/ChainGuardian/node_modules/bip39/types/index"' has no default export.

test/components/HDWallet.spec.tsx
jogamod: Do you mean test file or hdwallet?

src/renderer/services/Eth1ICGKeystore.ts
mpetrunic: interesting. I would swear  it works on another project. Well use `import {writeFileSync} from "fs"` then :)

src/renderer/services/Eth1ICGKeystore.ts
mpetrunic: ethereumjs-wallet is written in typescript so the library itself already contains typescript

src/renderer/services/Eth2HDWallet.ts
mpetrunic: oh, stupid libraries. Either enable syntetic default export in tsconfig or leave it like this

test/components/HDWallet.spec.tsx
mpetrunic: test file

mpetrunic:
@jogamod can you merge master?

mpetrunic:
and you have some lint issues

src/renderer/services/wallets/WalletInterfaces.ts
aostrun: Can we use something like this for the `ICGKeystore` as it will be easier to create interface implementations, problem is with the constructor signature  `new` ([here is the reason why](https://www.typescriptlang.org/docs/handbook/interfaces.html#difference-between-the-static-and-instance-sides-of-classes)).
```javascript
export interface ICGKeystore {
    decrypt(password: string): Keypair;
    changePassword(oldPassword: string, newPassword: string): void;
    destroy(): void;
    create(file: string, password: string, keypair: Keypair): ICGKeystore;
}

export interface ICGKeystoreFactory {
    new (file: string): ICGKeystore;
}
```

PR #27, opened 2019-10-15
Feature/action buttons, ButtonStandard fix

irubido:


.storybook/config.ts
```diff
@@ -10,3 +16,8 @@ function loadStories() {
 }
 
 configure(loadStories, module);
+
+
+export interface Interface {
+    new(aas: string): Interface
+}
```
mpetrunic: Think you should remove this

package.json
```diff
@@ -145,14 +146,13 @@
   },
   "dependencies": {
     "@chainsafe/eth2.0-types": "0.1.0",
+    "@storybook/addon-backgrounds": "^5.2.3",
```
mpetrunic: should be devDependency

src/renderer/components/Button/ButtonAction.tsx
```diff
@@ -0,0 +1,46 @@
+import * as React from "react";
+// import { ReactComponent as Arrow } from './Icon/Vectorbacktab-1.svg';
+import arrow from "./Icon/Backtab.svg";
+import add from "./Icon/Add.svg";
+import copyDefault from "./Icon/CopyDefault.svg";
+import copyAll from "./Icon/CopyAll.svg";
+import copied from "./Icon/Copied.svg";
+
+export interface IProps {
```
mpetrunic: Can you rename to IActionButttonProps

test/components/Button.spec.tsx
```diff
@@ -0,0 +1,25 @@
...
+            .toJSON();
+        expect(tree).toMatchSnapshot();
+    });
+    it("renders focused button", () => {
+        const tree = renderer
+            .create(<ButtonPrimitive focused={true}>Submit</ButtonPrimitive>)
+            .toJSON();
+        expect(tree).toMatchSnapshot();
+    });
+});
```
mpetrunic: I think it would be best to copy these test cases for all standard buttons.

PR #28, opened 2019-10-15
Create deposit transaction

MakMuftic:
resolves #9

PR #31, opened 2019-10-17
Create Account model

aostrun:
Account model implementation.


src/renderer/models/account.ts
```diff
@@ -0,0 +1,136 @@
+import {Keypair} from "@chainsafe/bls/lib/keypair";
+import {readdirSync, readFileSync} from "fs";
+import {ICGKeystore, ICGKeystoreFactory} from "../services/interfaces";
+
+export interface ISSZAccount {
```
mpetrunic: Rename it to just IAccount, it isn't ssz specific

src/renderer/models/account.ts
```diff
@@ -0,0 +1,136 @@
...
+export class CGAccount implements ISSZAccount {
+    public name: string;
+    public directory: string;
+    public sendStats: boolean;
+
+    private unlocked = false;
+    private validators: Keypair[] = [];
+    private keystoreTarget: ICGKeystoreFactory;
+
+    constructor(name: string, directory: string, sendStats = false, keystoreTarget: ICGKeystoreFactory) {
```
mpetrunic: - might be easier if constructor accepts `account: IAccount` as first param
- you should provide default KeystoreFactory

src/renderer/models/account.ts
```diff
@@ -0,0 +1,136 @@
...
+        const keystoreFiles: string[] = this.getKeystoreFiles();
+
+        const validatorAddresses: string[] = [];
+        keystoreFiles.forEach(file => {
+            const fileContents = readFileSync(this.directory + file);
+            const fileJSON = JSON.parse(fileContents.toString());
+            if (fileJSON.address) {
+                validatorAddresses.push(fileJSON.address);
+            }
+        });
```
mpetrunic: this will duplicate code. I would suggest something like:

```javascript
const keystorePaths = this.getKeystoreFiles();
keystorePaths.map(this.keystoreFactory.new).map(keystore => keystore,getAddress())
```

mpetrunic: make sure to handle parsing errors though

src/renderer/models/account.ts
```diff
@@ -0,0 +1,136 @@
...
+    /**
+     * returns all validator keypairs or throws if not unlocked
+     * @param password decryption password of the keystore
+     */
+    // eslint-disable-next-line
+    getValidators(password: string): Keypair[] {
+        // TODO: loop trough keystore files and unlock them
+        if (!this.unlocked) {
+            throw new Error("Keystore locked.");
+        }
```
mpetrunic: You should implement todo. Probably something similar as above comment

src/renderer/models/account.ts
```diff
@@ -0,0 +1,136 @@
...
+         * ? As there can be multiple keystore files there can also be
+         * ? different passwords that these keystores use, we need to
+         * ? define how are we going to handle these situations.
+         * * Currently, if any of the keystores matches the provided
+         * * password this method returns true.
+         */
+
+        const keystoreFiles = this.getKeystoreFiles();
+
+        for (const fileIdx in keystoreFiles) {
```
mpetrunic: As discussed, just take first keystore

mpetrunic: in future we can consider encrypting whole directory, not sure

src/renderer/models/account.ts
```diff
@@ -0,0 +1,136 @@
...
+
+        keystoreFiles.forEach(file => {
+            const filePath = this.directory + file;
+            const keystore = new this.keystoreTarget(filePath);
+
+            try {
+                const keypair = keystore.decrypt(password);
+
+                this.validators.push(keypair);
+                this.unlocked = true;
```
mpetrunic: you could probably create helkper function
```javascript
private isUnlocked() {
  return this.validators.length > 0;
}
```

src/renderer/services/db/api/repositories/account.ts
```diff
@@ -0,0 +1,16 @@
...
+import {CGAccount} from "../../../../models/account";
+import {encodeKey} from "../../schema";
+
+export class AccountRepository extends Repository<CGAccount> {
+    // Override get method to wrap deserialized data into CGAccount instance
+    public async get(id: Id): Promise<CGAccount | null> {
+        try {
+            const value = await this.db.get(encodeKey(this.bucket, id));
+            if (!value) return null;
+            return this.serializer.deserialize(value, this.type);
```
mpetrunic: Think you are missing the wrapping part? :)

src/renderer/services/interfaces.ts
```diff
@@ -1,5 +1,31 @@
...
+export interface ICGKeystore {
+    decrypt(password: string): Keypair;
+    changePassword(oldPassword: string, newPassword: string): void;
+    //deletes physical keystore file
+    destroy(): void;
+    create(file: string, password: string, keypair: Keypair): ICGKeystore;
+}
+
+export interface ICGKeystoreFactory {
+    new (file: string): ICGKeystore;
```
mpetrunic: maybe it would be nicer if this wan't constructor rather "create" method;
You could make if in test to return sino.stub()

test/test_keystores/keystore1.json
```diff
@@ -0,0 +1,19 @@
...
+        "kdfparams": {
+            "dklen": 32,
+            "salt": "2679d31db8b7be7cd8f9c59fbe9e4df8a1c3d9268a001782c4ba1798ee0fb916",
+            "n": 131072,
+            "r": 8,
+            "p": 1
+        },
+        "mac": "6f28f837c69cdc89b87a1b696d92b38e6bc60fa5b4ac166d65871afdabb2e6bb"
+    }
+}
```
mpetrunic: this should go in e2e

test/unit/Models/account.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+/*eslint-disable*/
+const CGKeystoreTest: ICGKeystoreFactory = class CGKeystoreTest implements ICGKeystore {
+    private file: string;
+    constructor(file: string) {
+        this.file = file;
+    }
+
+    decrypt(password: string): Keypair {
+        if (password === PRIMARY_KEYSTORE_PASSWORD) {
+            return Keypair.generate();
```
mpetrunic: this is faster:
https://github.com/NodeFactoryIo/ChainGuardian/pull/28/files#diff-f55bcd7f099136dc7b3589dd8b2728a1R6

test/unit/Models/account.spec.ts
```diff
@@ -0,0 +1,96 @@
+import {CGAccount} from "../../../src/renderer/models/account";
+import {ICGKeystore, ICGKeystoreFactory} from "../../../src/renderer/services/interfaces";
+import {Keypair} from "@chainsafe/bls/lib/keypair";
+
+// Passwords for keystores 1 & 2
+const PRIMARY_KEYSTORE_PASSWORD = "chainGuardianPass";
+
+/*eslint-disable*/
+const CGKeystoreTest: ICGKeystoreFactory = class CGKeystoreTest implements ICGKeystore {
```
mpetrunic: I think this should return sinon.stubs

test/unit/Models/account.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+        throw new Error("Method not implemented.");
+    }
+    create(file: string, password: string, keypair: Keypair): ICGKeystore {
+        throw new Error("Method not implemented.");
+    }
+};
+/*eslint-enable*/
+
+function createTestAccount(): CGAccount {
+    return new CGAccount("Test Account", "./test/test_keystores/", false, CGKeystoreTest);
```
mpetrunic: you should not use real keystore implementation. You keystore should be stubbed with sinon so you can simulate errors and stuff and we won't have to wait for decryption  in unit tests

mpetrunic:
@aostrun could you fix those errors?

src/renderer/models/account.ts
```diff
@@ -0,0 +1,141 @@
...
+
+        return validatorAddresses;
+    }
+
+    /**
+     * returns all validator keypairs or throws if not unlocked
+     * @param password decryption password of the keystore
+     */
+    // eslint-disable-next-line
+    getValidators(password: string): Keypair[] {
```
mpetrunic: why do you have password param?

src/renderer/services/Eth1ICGKeystore.ts
```diff
@@ -51,10 +53,12 @@ export class Eth1ICGKeystore implements ICGKeystore {
...
     decrypt(password: string): Keypair {
-        const privateKeyString = eth1WalletProvider
-            .fromV3(JSON.stringify(this.keystore), password)
-            .getPrivateKeyString();
+        const keystore = eth1WalletProvider
+            .fromV3(JSON.stringify(this.keystore), password);
+        const privateKeyString = keystore.getPrivateKeyString();
         const priv = PrivateKey.fromHexString(privateKeyString);
+        // Collect address from keystore
+        this.address = keystore.getAddressString();
```
mpetrunic: you should be able to fetch address without unlocking

src/renderer/services/Eth1ICGKeystore.ts
```diff
@@ -9,6 +9,8 @@ export class Eth1ICGKeystore implements ICGKeystore {
     keystore: IV3Keystore;
     file: string;
 
+    private address = "";
```
mpetrunic: this is redundant

src/renderer/models/account.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * should return addresses from validator keystores in account directory
+     */
+    getValidatorsAddresses(): string[] {
+        // Loop trough files in account directory
+        const keystoreFiles: string[] = this.getKeystoreFiles();
+
+        const validatorAddresses: string[] = keystoreFiles
+            .map(file => new this.keystoreTarget(file))
```
mpetrunic: keystore.keystore.address holds the address

src/renderer/models/account.ts
aostrun: It is unnecessary, I agree.

src/renderer/services/Eth1ICGKeystore.ts
aostrun: fixed

src/renderer/models/account.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    /**
+     * should return addresses from validator keystores in account directory
+     */
+    public getValidatorsAddresses(): string[] {
+        // Loop trough files in account directory
+        const keystoreFiles: string[] = this.getKeystoreFiles();
+
+        const validatorAddresses: string[] = keystoreFiles
+            .map(file => new this.keystoreTarget(file))
```
mpetrunic: I think this can throw exception if user puts invalid keystore in app directory (manual import) so you should filter our keystores tha throws expetion. It probably makes sense and it would reduce code duplication that `getKEystoreFiles` returns keystore instances (already filtered)

src/renderer/services/db/api/repositories/account.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export class AccountRepository extends Repository<CGAccount> {
+    // Override get method to wrap deserialized data into CGAccount instance
+    public async get(id: Id): Promise<CGAccount | null> {
+        try {
+            const value = await this.db.get(encodeKey(this.bucket, id));
+            if (!value) return null;
+            const {name, directory, sendStats} = this.serializer.deserialize(value, this.type);
+
+            return new CGAccount({name, directory, sendStats});
```
mpetrunic: you should convert buffer to string here. When deserializez bytes<n> types are Buffers

src/renderer/models/ssz/types.ts
```diff
@@ -1,9 +1,12 @@
-import {SimpleContainerType} from "@chainsafe/ssz";
+import {SimpleContainerType, Type} from "@chainsafe/ssz";
 
-export const Account = (): SimpleContainerType => ({
+// @ts-ignore
```
mpetrunic: why ts ignore?

src/renderer/models/ssz/types.ts
```diff
@@ -1,9 +1,12 @@
...
+export const Account: SimpleContainerType = {
+    // @ts-ignore
+    type: Type.container,
```
mpetrunic: you don't need type though, think AnyContainerType should suffice

src/renderer/services/db/api/repositories/account.ts
```diff
@@ -1,18 +1,39 @@
...
     }
+
+    public async set(id: Id, value: CGAccount): Promise<void> {
+        const parsedAccount = {
+            name: value.name,
+            directory: value.directory,
+            sendStats: value.sendStats
+        };
+
+        //@ts-ignore
```
mpetrunic: You can remove this if you assign type to parsedAccount

src/renderer/models/ssz/types.ts
aostrun: We need it for the JSON serializer as we want to extract only specific attributes from the object and we need to know if the object passed to the serializer is actually a container or a type literal.
Check the `serializers/json.ts` to see this in action.

But yeah with the AnyContainerType we don't need the ts-ignore.

src/renderer/services/db/api/repositories/account.ts
aostrun: Actually we don't even need this override of the set method.

src/renderer/models/ssz/types.ts
aostrun: Answered at the comment bellow.

src/renderer/models/ssz/types.ts
mpetrunic: I think in ssz, container is always if it has key "fields"

aostrun: I managed to remove the ts-ignore but I'm still using `type: Type.container` (I switched to `AnyContainerType` as you suggested). I can remove it and instead check if `field` exist and do the container check that way, but the current implementation also works so it is down to personal preference, what do you think? @mpetrunic

mpetrunic: I would suggest checking if fields prop exists just to be compliant with how ssz is implemented

mpetrunic: btw you have some lint errors

aostrun: Lint errors are frontend related and I didn't want to touch things that I'm not assigned to.
Errors appeared as a result of the `@typescript-eslint/explicit-member-accessibility`. 
I can fix it in this PR if you are ok with me touching frontend code?

mpetrunic: you have to or we cannot merge :)

aostrun: oh, ok 😅

mpetrunic:
@aostrun merge master, you have conflicts

PR #50, opened 2019-10-29
Feature/switch to ethersjs

### LINKED ISSUE #43: Switch to ethers.js
Since we will have to fetch and parse events, we have already pulled most of dependencies of ethers.js so we should remove ethereumjs stuff and replace with ethers.js methods

MakMuftic:
resolves #43


src/renderer/services/utils/crypto-utils.ts
```diff
@@ -11,17 +12,59 @@ export function toHexString(data: string | Buffer | BN): string {
 }
 
 /**
- * Provides methods to convert amount defined in eth to wei / gwei.
+ * Provides methods to convert amounts from currencies defines as @{Currency}.
  */
 export const EthConverter = {
```
MakMuftic: @mpetrunic We still need conversion util because ethers util methods only cover _eth -> wei_ conversion. Maybe I can fully extract this utility and publish it over npm as lib?

PR #55, opened 2019-10-30
Fix/ step navigation onTab prop

### LINKED ISSUE #49: Tabnavigation must be clickable
- onClick handle should return id of clicked tab

irubido:
resolves #49


src/renderer/components/TabNavigation/TabNavigation.stories.tsx
```diff
@@ -21,6 +24,6 @@ storiesOf('Tab Navigation', module).add('Tab Navigation', () => {
     ];
     const array = object("steps", d);
     return <div >
-            <TabNavigation tabs={array} current={selected}/>
+            <TabNavigation onTab={()=>setSelected(selected)} tabs={array} current={selected}/>
```
mpetrunic: should be
```javascript
<TabNavigation onTab={(tab)=>setSelected(tab)} tabs={array} current={selected}/>
```

PR #60, opened 2019-10-31
Feature - Http metrics

aostrun:
Implemented a ~~method decorator~~ `method wrapper`  that allows us to track metrics of any method.
The ~~decorator~~ `wrapper` is very flexible and should allow us to use it across various classes.

~~Usage of the decorator can be found in the `src/renderer/services/api/http/httpClient.ts`.~~

Usage can be found in the `src/renderer/services/api/BeaconAPIClientMetrics.ts`.


src/renderer/services/api/http/httpClient.ts
```diff
@@ -13,6 +14,7 @@ export class HttpClient {
      * Method that handles GET
      * @param url endpoint url
      */
+    @trackMetrics()
```
mpetrunic: I would rather put this on eth2 client than on every http api call

tsconfig.json
```diff
@@ -57,7 +57,7 @@
     // "inlineSources": true,                 /* Emit the source alongside the sourcemaps within a single file; requires '--inlineSourceMap' or '--sourceMap' to be set. */
 
     /* Experimental Options */
-    // "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
+    "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
```
mpetrunic: can babel build this?

src/renderer/services/db/api/database.ts
```diff
@@ -1,13 +1,25 @@
...
         super(opts);
         this.account = new AccountRepository(this.config, this.db);
+        this.httpMetrics = new HttpMetricsRepository(this.config, this.db);
     }
 
+}
+
+export function initDB(opts: IDatabaseApiOptions): void{
+    database = new CGDatabase(opts);
+}
```
mpetrunic: We don't really wanna do this. Better to do it using events. Like create som EventBus singleton where you post events and some MetricsCollector that subscribes to those events on start.

mpetrunic:
@aostrun we messed this up. I totally forgot we will have multiple beacon node clients. And they have to be tracked separately which we cannot do using decorators. I think we need to fall back to wrapper class around BeaconApiClient

src/renderer/services/metrics/Metrics.ts
```diff
@@ -0,0 +1,25 @@
...
+        const currentDate = new Date();
+        const metrics = new Metrics({
+            metric: end - start,
+            method: functionName,
+            date: currentDate.toISOString(),
+        });
+        // YYYY-MM-DD-methodName-itearation
+        const id = `${currentDate.toISOString()}-${functionName}`;
+        // Save the metrics to the DB based on the provided Bucket
+        db.metrics.set(
```
mpetrunic: create method on metrics repo that forms key from given date, function name and beacon client id(introduce that :) )

src/renderer/services/api/BeaconAPIClient.ts
```diff
@@ -34,6 +36,7 @@ export class BeaconAPIClient implements IBeaconAPIClient {
         }
         this.options = options;
         this.httpClient = new HttpClient(options.urlPrefix);
+        this.instanceId = v4();
```
mpetrunic: should come from options, ether url or user given name

src/renderer/services/db/api/repositories/metrics.ts
```diff
@@ -11,3 +11,8 @@ export class MetricsRepository extends BulkRepository<Metrics> {
         super(config, db, JSONSerializer, Bucket.generalMetrics, SZMetrics);
     }
 }
+
+
+export function generateMetricsKey(date: Date, method: string, instanceId: string): string{
+    return `${date.toISOString()}-${method}-${instanceId}`;
```
mpetrunic: I think because of searching, date must be last param

src/renderer/services/db/api/repositories/metrics.ts
```diff
@@ -14,5 +14,5 @@ export class MetricsRepository extends BulkRepository<Metrics> {
 
 
 export function generateMetricsKey(date: Date, method: string, instanceId: string): string{
-    return `${date.toISOString()}-${method}-${instanceId}`;
+    return `${instanceId}-${date.toISOString()}-${method}`;
```
mpetrunic: this date format doesn't really fit, we need something like `yyyymmddhh`

aostrun: Are you sure about this? It's probably better to stick with one of the date representation standards and ISO (YYYY-MM-DDTHH:mm:ss.sssZ) is pretty good for our case as it can be queried in multiple ways:
- filter by year (you can also set the year range)
```
gte: '<instanceId><selected-year>'
lte: '<instanceId><selected-year>~'
```

- filter by month (month range also available)
```
gte: '<instanceId><selected-year>-<selected-month>'
lte: '<instanceId><selected-year>-<selected-month>~'
```

- filter by minute
```
gte: '<instanceId><selected-year>-<selected-month>-<selected-day>T<selected-hour>:<selected-minute>'
lte: '<instanceId><selected-year>-<selected-month>-<selected-day>T<selected-hour>:<selected-minute>~'
```
- etc.

So ISO is basically very similar to what you are proposing but it is standardized and quite flexible.

mpetrunic: valid arguments, lrts keep it this way. my argument was that it makes filtering methods impossible but we have to filter mannually anyways

aostrun: Yeah, we can not query by the last parameter (methodName) in any case. 
We could possibly restrict the query for the method name by querying the events that happened in the last x days/months/years and then filter the result to extract entries that reference desired method.

mpetrunic: I think it's better to save it under different key but we don't need it for now (filtering by method)

src/renderer/services/metrics/Metrics.ts
```diff
@@ -0,0 +1,33 @@
...
+import {generateMetricsKey} from "../db/api/repositories/metrics";
+
+export interface IMetricsParams {
+    instanceId: string,
+    function: Function,
+    functionName: string,
+    bucket: Bucket,
+}
+
+export function measureExecution<T>(instance: any, params: IMetricsParams, ...args: any[]): T{
```
mpetrunic: there is a neat typescript trick for this.

```javascript

export function measureExecution<ReturnType>(this: BeaconClientMetrics, params: IMetricsParams, ...args: unknown[]): ReturnType;

//BeaconCLientMetrics.ts
this.measureExecution: typeof measureExecution = measureExecution;
```

src/renderer/services/metrics/Metrics.ts
```diff
@@ -0,0 +1,33 @@
...
+    const db = getDB();
```
mpetrunic: replace with this.db (db should be protected in "this")

src/renderer/services/metrics/Metrics.ts
```diff
@@ -0,0 +1,33 @@
...
+    const end = performance.now();
+
+    const currentDate = new Date();
+    const metrics = new Metrics({
+        metric: end - start,
+        method: params.functionName,
+        date: currentDate.toISOString(),
+    });
+        // YYYY-MM-DD-methodName-instanceId
+    const id = generateMetricsKey(currentDate, params.functionName, params.instanceId);
```
mpetrunic: move this to `db.metrics.store(metrics);`

src/renderer/services/api/BeaconAPIClientMetrics.ts
```diff
@@ -0,0 +1,144 @@
...
+    IndexedAttestation,
+    Shard,
+    uint8
+} from "@chainsafe/eth2.0-types";
+import {IBeaconApiClientOptions, ForkInformation, Syncing} from "./interface";
+import {Bucket} from "../db/schema";
+
+export class BeaconAPIClientMetrics extends BeaconAPIClient {
+
+    public constructor(options: IBeaconApiClientOptions) {
```
mpetrunic: you should receive BeaconAPIClient instance and db instance

src/renderer/services/api/BeaconAPIClientMetrics.ts
```diff
@@ -0,0 +1,144 @@
...
+    Slot,
+    BeaconBlock,
+    IndexedAttestation,
+    Shard,
+    uint8
+} from "@chainsafe/eth2.0-types";
+import {IBeaconApiClientOptions, ForkInformation, Syncing} from "./interface";
+import {Bucket} from "../db/schema";
+
+export class BeaconAPIClientMetrics extends BeaconAPIClient {
```
mpetrunic: implements IBeaconAPIClient not extends

test/e2e/Services/metrics/BeaconMetrics.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+ * *******************
+ * Mock axios requests
+ * *******************
+ */
+
+mock.onGet(FETCH_NODE_VERSION).reply(200, "1");
+mock.onGet(FETCH_GENESIS_TIME).reply(200, 2);
+
+
+describe("Http Metrics", () => {
```
mpetrunic: move to unit tests, stub out db and beacon client

src/renderer/services/api/BeaconAPIClientMetrics.ts
aostrun: I tried to create a _plug & play_ metrics service that would allow us to quickly wrap a class and extend it with metrics functionalities.
At first, I didn't want to create a `BeaconAPIClient` implementation that would include metrics as we would need to manually update every occurrence of the "old" `BeaconAPIClient` and as we are adding dependencies to the client (`BeaconAPIClient` instance and `DB` instance) we would need to inject them manually.

I tried to implement metrics in such a way that would allow us to add metrics functionality without any hassle around additional dependencies, that's why I'm extending existing implementations (wrapping existing implementation) and not implementing custom implementation. That's why I'm currently using a "global" DB instance (`getDB()`) and whoever is going to use this client doesn't have to bother with injecting additional dependencies.

Of course, this all makes sense if we are going to use this metrics service later on for other classes, but if we are going to use it only for the `BeaconAPIClient`  we could definitely do it as you suggested (creating a custom implementation by wrapping existing instance of the client and not extending one).

src/renderer/services/api/BeaconAPIClientMetrics.ts
aostrun: Refer to the comment above.

src/renderer/services/api/BeaconAPIClientMetrics.ts
mpetrunic: we try to avoid singleton database as it makes testing and controlling of database wierd, especially since we operate in two threads so  by all means all dependencies need to be injected in constructor so they can be easily stubed in tests and swapped.

metrics function can be typed so it has to be binded by function which has property db. But to keep it simple for now, allow it to only be binded by BeaconApiClientMetrics instance.

This should allow you to write real unit tests and not worry about api or database implementation like you do now.

PR #63, opened 2019-11-07
Feature/deposit contract 0.9.0

### LINKED ISSUE #51: Update deposit contract to 0.9.0
- https://github.com/ethereum/eth2.0-specs/pull/1362

jogamod:
From v0.8.3 to v0.9.0

resolves #51


src/renderer/services/deposit/deposit-tx.ts
```diff
@@ -1,9 +1,11 @@
 import {bytes, DepositData} from "@chainsafe/eth2.0-types";
+import {config} from "@chainsafe/eth2.0-config/lib/presets/mainnet";
```
mpetrunic: didn't see that in original implementation. Config should be param so we can support testnets

PR #71, opened 2019-11-12
Feature/user deposit

jogamod:
1. Retroactive search for user deposits and filtering them with the certain public key
2. Listening for deposit events


src/renderer/services/deposit/ethers.ts
```diff
@@ -0,0 +1,82 @@
...
+export class EthersNotifier {
+    private provider: ethers.providers.Web3Provider;
+    private depositContractAddress: string;
+    private config: IBeaconConfig;
+    private signingKey: Keypair;
+
+    public constructor(
+        provider: ethers.providers.Web3Provider,
+        depositContractAddress: string,
+        config: IBeaconConfig,
```
mpetrunic: I was thinking, since we will support various testnets,
you could create interface:
```javascript
export interface NetworkConfig {
   config: IBeaconConfig;
   contract: {
      address: string,
      bytecode: string,
      deployedAtBlock: number
   }
}
```

src/renderer/services/deposit/ethers.ts
```diff
@@ -0,0 +1,82 @@
...
+    public depositEventListener(timeout: number): Promise<BN>{
+        return new Promise((resolve, reject) => {
+            const contract = new Contract(this.depositContractAddress, DepositContract.abi, this.provider);
+            const filter = contract.filters.DepositEvent(null);
+    
+            // Listen for our filtered results
+            contract.on(filter, (pubkey, withdrawalCredentials, amount) => {
+                if (pubkey === this.signingKey.publicKey.toHexString()) {
+                    const amountGwei = deserialize(Buffer.from(amount.slice(2), "hex"), this.config.types.Gwei) as Gwei;
+                    console.log("Received " + amountGwei.toString() + " gwei from " + pubkey);
```
mpetrunic: remove console

src/renderer/services/deposit/ethers.ts
```diff
@@ -0,0 +1,82 @@
...
+            const timer = setTimeout(() => {
+                contract.removeAllListeners(DEPOSIT_EVENT);
+                reject(new Error("Timeout waiting for deposit event"));
+            }, timeout);
+        });
+    }
+
+    public async checkUserDepositAmount(): Promise<bool> {
+        const filter = {
+            fromBlock: 0,
```
mpetrunic: use block from new config because this won't work in infura

src/renderer/services/deposit/ethers.ts
```diff
@@ -0,0 +1,82 @@
...
+            address: this.depositContractAddress
+        };
+    
+        const logs = await this.provider.getLogs(filter);
+        const amountSum = new BN(0);
+    
+        logs.forEach((log: any) => {
+            const data = utils.defaultAbiCoder.decode(
+                DepositContract.abi[0].inputs.map((parameter: any) => parameter.type),
```
mpetrunic: think ethers has `ParamType` exported

PR #78, opened 2019-11-21
keystore v4

### LINKED ISSUE #77: Replace v3 keystore with v4
bls keystore implementation: https://github.com/nodefactoryio/bls-keystore

jogamod:
resolves #77


package.json
```diff
@@ -39,8 +39,8 @@
     "lint:fix": "yarn run lint --fix",
     "check-types": "tsc --noEmit",
     "test": "yarn run test:unit && yarn run test:components && yarn run test:e2e && yarn run test:screens",
-    "test:unit": "yarn run jest --coverage \"test/unit\"",
-    "test:e2e": "yarn run jest \"test/e2e\"",
+    "test:unit": "ELECTRON=1 yarn run jest --coverage \"test/unit\"",
+    "test:e2e": "ELECTRON=1 yarn run jest \"test/e2e\"",
```
mpetrunic: we probably need that env in start and start-dev?

package.json
```diff
@@ -170,7 +170,9 @@
     "@chainsafe/eth2.0-types": "0.1.0",
     "@chainsafe/ssz": "^0.5.1",
     "@hapi/joi": "^16.1.7",
+    "@nodefactory/bls-keystore": "^1.0.0-beta7",
     "axios": "0.19.0",
+    "bcrypto": "^4.2.8",
```
mpetrunic: Don't think we need it (we have it already trough bls-keystore)

src/renderer/models/account.ts
```diff
@@ -18,7 +18,7 @@ export class CGAccount implements IAccount {
 
     public constructor(
         account: IAccount,
-        keystoreTarget: ICGKeystoreFactory = Eth1KeystoreFactory
+        keystoreTarget: ICGKeystoreFactory = EthKeystoreFactory
```
mpetrunic: Eth2KeystoreFactory should be appropriate name

mpetrunic: or we can change eth1 to v3 and eth2 to v4 keystore naming

mpetrunic: I like second solution

src/renderer/services/keystore/eth2.ts
```diff
@@ -58,18 +49,17 @@ export class Eth1Keystore implements ICGKeystore {
...
+            writeFileSync(this.file, keystore.toJSON());
             this.keystore = keystore;
         } catch (err) {
             throw new Error(`Failed to write to ${this.file}: ${err}`);
         }
     }
 
     public getAddress(): string {
-        return this.keystore.address;
+        return this.keystore.pubkey.toString();
```
mpetrunic: I would maybe expose another method getPublicKey and make this method convert public key to address using bech32

test/.npmrc
```diff
@@ -0,0 +1 @@
+openssl_dir = /usr/local/ssl
```
mpetrunic: we can remove this

test/unit/keystore/eth2.spec.ts
```diff
@@ -31,53 +30,57 @@ describe("Eth1ICGKeystore", () => {
...
+        eth2Keystore.decrypt(password);
         expect(writeStub.calledOnce).toEqual(true);
         expect(readStub.calledOnce).toEqual(true);
+
     });
 
+
     it("should decrypt", async () => {
-        const keypair = await eth1Keystore.decrypt(password);
+        const keypair = await eth2Keystore.decrypt(password);
```
mpetrunic: That slipped me with first version of keystore. Could you make test independent of each other using beforeEach or something

src/renderer/services/keystore/eth2.ts
```diff
@@ -59,6 +61,11 @@ export class Eth2Keystore implements ICGKeystore {
     }
 
     public getAddress(): string {
+        const words = bech32.toWords(Buffer.from(this.keystore.pubkey, "utf-8"));
```
mpetrunic: Thinks that's wrong.
Buffer.from(this.getPublicKey(), "hex")?

src/renderer/services/keystore/eth2.ts
```diff
@@ -59,6 +61,11 @@ export class Eth2Keystore implements ICGKeystore {
...
+        return bech32.encode(ETH2_ADDRESS_PREFIX, words);
+    }
+
+    public getPublicKey(): string {
         return this.keystore.pubkey.toString();
```
mpetrunic: isn't this suppose to be .toString("hex")

test/unit/keystore/v4.spec.ts
```diff
@@ -32,55 +53,37 @@ describe("Eth1ICGKeystore", () => {
 
 
     it("should create keystore", async () => {
```
mpetrunic: This test is probably useless. It would probably make sense to have something like V4Keystore.fromJSON(JSON.stringify(example)) in beforeach. And this first test should test creating of new keystore.

It would be a lot more cleaner

src/renderer/services/keystore/eth2.ts
jogamod: I just figured out that `this.keystore.pubkey` is already a hex string

jogamod: there is not toString method on string. I will return just pubkey, since it is already a string

PR #81, opened 2019-11-22
Create password container

MakMuftic:


src/renderer/containers/Onboard/CreatePassword/CreatePassword.tsx
```diff
@@ -0,0 +1,87 @@
...
+    input: string;
+    confirmInput: string;
+    valid?: boolean;
+    validationMessage: string;
+    confirmed?: boolean;
+}
+
+const SEPARATOR = "and";
+
+export default class CreatePassword extends Component<{ history: History }, {}> {
```
mpetrunic: - no export default, it messes up import typehinting.
- `extends Component<Pick<RouteComponentProps, "history">>`
- CreatePasswordContainer

src/renderer/containers/Onboard/CreatePassword/CreatePassword.tsx
```diff
@@ -0,0 +1,87 @@
...
+            {
+                inputId:"inputPassword",
+                onChange: this.handleChange,
+                placeholder: "Enter password",
+                valid: this.state.valid,
+                errorMessage: this.state.validationMessage
+            },
+            {
+                inputId:"confirmPassword",
+                onChange: this.compareInput,
```
mpetrunic: why not use joi validation for confirm?

src/renderer/containers/Onboard/CreatePassword/CreatePassword.tsx
MakMuftic: Regarding the last point, is it really a separate container, because it is part of OnboardContainer?

src/renderer/components/MultipleInputVertical/MultipleInputVertical.stories.tsx
```diff
@@ -0,0 +1,22 @@
+import { storiesOf } from '@storybook/react';
+import { withKnobs } from '@storybook/addon-knobs';
+import { MultipleInputVertical } from './MultipleInputVertical';
+import { IInputFormProps } from '../Input/InputForm';
+import * as React from 'react';
```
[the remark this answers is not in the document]
MakMuftic: If this is regarding ```import * as React from 'react';``` I am pretty sure that it needs to be there (open any ```.stories``` file for component and you will see the same import).

I saw multiple unused dep trough code (...see: ```StepNavigation.stories.tsx```), can I do this code clean up in this branch / or open a new one for that?

src/renderer/containers/Onboard/CreatePassword/CreatePassword.tsx
MakMuftic: @mpetrunic

mpetrunic: It handles inputs and navigation and everything so it can't be component

src/renderer/components/MultipleInputVertical/MultipleInputVertical.stories.tsx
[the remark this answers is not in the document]
mpetrunic: leave it, wasn't aware that it doesn't work without it

src/renderer/containers/Onboard/CreatePassword/CreatePassword.tsx
MakMuftic: I could use Joi validation as ``` Joi.string().valid(this.state.input).validate(e.currentTarget.value) ```
but it wouldn't change the need for saving this confirmation password string in the state (_there are few input flows that would brake otherwise_). 
That is why I decided to go with a simple comparison because I think it is overkill to use Joi to see if two values are the same? 
I can show you tomorrow which flows are problematic.

mpetrunic: but if you have state like this:
```
{
   form: {
    password: string,
    confirm: string
}
}
```
you can pass entire form to joi and use joi.ref for comparing

src/renderer/containers/Onboard/CreatePassword/CreatePasswordContainer.tsx
```diff
@@ -0,0 +1,101 @@
...
+        const selector = (e.currentTarget.id === "inputPassword") ? "password" : "confirm";
+        const localState = this.state;
+        localState[selector] = e.currentTarget.value;
+        this.setState({[selector]: e.currentTarget.value});
+        // validate state
+        const validation = passwordFormSchema.validate(localState, {abortEarly: false});
+        // generate error messages
+        const m: {password?: string, confirm?: string} = {password: "", confirm: ""};
+        if (validation.error) {
+            const errors = joiValidationToErrorDetailsMessages(validation.error);
```
mpetrunic: Just display those errors, why do you have all code below?

We cannot merge this method as it's way too huge and unmaintainable. Just make it as simple as possible, we will worry about UX later on,

src/renderer/containers/Onboard/CreatePassword/CreatePasswordContainer.tsx
```diff
@@ -0,0 +1,101 @@
...
+        return (typeof error === "undefined") ? error : (error === "");
+    }
+
+    private handleChange = (e: React.FormEvent<HTMLInputElement>): void => {
+        const selector = (e.currentTarget.id === "inputPassword") ? "password" : "confirm";
+        const localState = this.state;
+        localState[selector] = e.currentTarget.value;
+        this.setState({[selector]: e.currentTarget.value});
+        // validate state
+        const validation = passwordFormSchema.validate(localState, {abortEarly: false});
```
mpetrunic: abortEarly: true

src/renderer/containers/Onboard/CreatePassword/validation.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const passwordFormSchema = Joi.object().keys({
+    password: Joi.password()
+        .min(6)
+        .max(12)
+        .numOfLower(1)
+        .numOfUpper(1)
+        .numOfNumbers(1)
+        .numOfSigns(1)
+        .required()
```
mpetrunic: required should be first check

src/renderer/containers/Onboard/CreatePassword/validation.ts
```diff
@@ -0,0 +1,27 @@
...
+import {Joi} from "../../../services/validation";
+
+export const passwordFormSchema = Joi.object().keys({
+    password: Joi.password()
+        .min(6)
+        .max(12)
+        .numOfLower(1)
+        .numOfUpper(1)
+        .numOfNumbers(1)
+        .numOfSigns(1)
```
mpetrunic: You could solve it all with this regex: 
`"^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)(?=.*[@$!%*?&])[A-Za-z\d@$!%*?&]{6,12}$"`

mpetrunic: but leave it as is.

src/renderer/containers/Onboard/CreatePassword/validation.ts
```diff
@@ -0,0 +1,27 @@
...
+        .required()
+        .messages({
+            "string.min": "Password must be at least 6 characters long",
+            "string.max": "Password must be at most 12 characters long",
+            "complexity.upper": "1 uppercase",
+            "complexity.lower": "1 lowercase",
+            "complexity.numbers": "1 numeric",
+            "complexity.signs": "1 special"
```
mpetrunic: just set all complexity messages to same message so you can abort early

PR #90, opened 2019-12-05
Balance graph draft

irubido:


src/renderer/components/BalanceGraph/BalanceGraph.stories.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { storiesOf } from '@storybook/react';
+import { withKnobs, text, boolean, number, select } from '@storybook/addon-knobs';
+import { BalanceGraph } from './BalanceGraph';
+import { IntervalEnum } from './BalanceGraph';
+
+storiesOf('BalanceGraph', module).add('BalanceGraph', () => {
+    
+    return <BalanceGraph 
+    defaultInterval={IntervalEnum.DAY}
+    onOptionClick={()=>{console.log("")}}
```
mpetrunic: remove console log as default return haha

src/renderer/components/BalanceGraph/BalanceGraph.stories.tsx
```diff
@@ -0,0 +1,20 @@
...
+    getData={()=>{
+        return(
+            [2356,3213,8934,7924,7924,1245,3456,5673,2124,
```
mpetrunic: I just realized, since it's balance graph and balance is in gwei, those are gonna be huge numbers. It's better to use bn.js.
Can recharts render bn.js?

src/renderer/components/BalanceGraph/BalanceGraph.tsx
```diff
@@ -0,0 +1,158 @@
...
+            case IntervalEnum.HOUR: setXAxis(dataValueArray, IntervalEnum.HOUR);
+                break;
+            case IntervalEnum.DAY: setXAxis(dataValueArray, IntervalEnum.DAY);
+                break;
+            case IntervalEnum.WEEK: setXAxis(dataValueArray, IntervalEnum.WEEK);
+                break;
+            case IntervalEnum.MONTH: setXAxis(dataValueArray, IntervalEnum.MONTH);
+                break;
+            case IntervalEnum.YEAR: setXAxis(dataValueArray, IntervalEnum.YEAR);
+                break;
```
mpetrunic: do you even need thos switch case. You could just do setXAxis(dataValueArray, interval)

src/renderer/components/BalanceGraph/BalanceGraph.tsx
```diff
@@ -0,0 +1,158 @@
...
+                    });
+                }
+                break;
+            }
+        }
+        setData(dataArray);
+    };
+
+    useEffect(()=>{
+        const dataValueArray = props.getData();
```
mpetrunic: props.getData(interval, 10)

src/renderer/components/BalanceGraph/BalanceGraph.tsx
```diff
@@ -0,0 +1,158 @@
...
+                    >1D</div>
+                    <div onClick={(): void=>{handleOptionClick(IntervalEnum.WEEK);}} 
+                        className={`graph-option ${(interval===IntervalEnum.WEEK ? "selected" : "")}`}
+                    >1W</div>
+                    <div onClick={(): void=>{handleOptionClick(IntervalEnum.MONTH);}} 
+                        className={`graph-option ${(interval===IntervalEnum.MONTH ? "selected" : "")}`}
+                    >1M</div>
+                    <div onClick={(): void=>{handleOptionClick(IntervalEnum.YEAR);}} 
+                        className={`graph-option ${(interval===IntervalEnum.YEAR ? "selected" : "")}`}
+                    >1Y</div>
```
mpetrunic: extract those in method like renderGraphIntervalOption(intervalEnum.YEAR, "1Y")

src/renderer/components/BalanceGraph/BalanceGraph.tsx
```diff
@@ -0,0 +1,158 @@
...
+        const dateToday = new Date();
+        switch (time) {
+            case IntervalEnum.HOUR: {
+                const minute = dateToday.getMinutes();
+                for (let i = 5; i >=0; i--) {
+                    dataArray.push({
+                        name: ((minute-10*i)<0 ? 60+(minute-10*i) : (minute-10*i)) + "'",
+                        value: array[i]
+                    });
+                }
```
mpetrunic: separate each case as function in another file. like normalizeDataPoints(array, interval) and have normalize function for each interval.

src/renderer/components/BalanceGraph/BalanceGraph.tsx
```diff
@@ -0,0 +1,158 @@
...
+    const [data, setData] = useState<Array<object>>([]);
+    const [interval, setInterval] = useState<IntervalEnum>(IntervalEnum.WEEK);
+
+    const formatDate = (date: string): string=>{
+        const month = date.slice(4,7).toUpperCase();
+        const day = date.slice(8,10);
+        return month + " " + day;
+    };
+
+    const setXAxis = (array: Promise<number[]>, time: IntervalEnum): void=>{
```
mpetrunic: this should not be promise, resolve that promise in initializator effect

src/renderer/components/BalanceGraph/BalanceGraph.tsx
irubido: Should "normalizeDataPoints" function have "interval" input, isn't it the same then as parent function "setXAxis = (array: number[], interval: IntervalEnum)", should I just separate setXAxis to another file with all cases?

mpetrunic: I would keep setXAxis because it's related to component, in other graphs you might setYAxis with normalized points :stuck_out_tongue:

src/renderer/components/BalanceGraph/BalanceGraph.tsx
```diff
@@ -56,92 +56,64 @@ export const BalanceGraph: React.FunctionComponent<IBalanceGraphProps> = (props:
...
     useEffect(()=>{
-    
         clearInterval(refreshInterval);
-        console.log("ON MOUNT: " + refreshInterval);
 
         const timeNow = new Date();
-        console.log(timeNow);
+        const msUntil = (60-(timeNow.getSeconds()))*1000;
+        const minUntil = ((60-(timeNow.getMinutes()-1))*1000*60)+msUntil;
+        const hourUntil = ((24-(timeNow.getHours()-1))*1000*60*60)+minUntil;
```
mpetrunic: I think it's better to save lastChecked time, and just check every minute

src/renderer/components/BalanceGraph/BalanceGraph.tsx
```diff
@@ -56,92 +56,64 @@ export const BalanceGraph: React.FunctionComponent<IBalanceGraphProps> = (props:
...
                 const intervalHandler = (): void =>{
                     awaitData();
                     const localRefreshInterval = window.setInterval(awaitData, 60000);
                     setRefreshInterval(localRefreshInterval);
                 };
-
-                window.setTimeout(intervalHandler,(60-seconds)*1000);
-
-                console.log("Case HOUR, refreshInterval on case: " + refreshInterval);
+                window.setTimeout(intervalHandler,msUntil);
```
mpetrunic: there is method called setInterval: https://www.w3schools.com/jsref/met_win_setinterval.asp

It will return reference, keep this reference and on unMounting of component clearInterval

src/renderer/components/BalanceGraph/BalanceGraph.tsx
```diff
@@ -0,0 +1,180 @@
...
+    //HOUR-last 60 minutes - 60 values
+    //DAY-last 24 hours - 24 values
+    //WEEK-last 7 days - 7 values
+    //MONTH-last 12 days - 12 values
+    //YEAR-last 12 months- 12 values
+}
+
+export const BalanceGraph: React.FunctionComponent<IBalanceGraphProps> = (props: IBalanceGraphProps) => {
+    const [data, setData] = useState<Array<object>>([]);
+    const [interval, setInterval] = useState<IntervalEnum>(props.defaultInterval);
```
mpetrunic: rename to intervalOption

src/renderer/components/BalanceGraph/BalanceGraph.tsx
```diff
@@ -0,0 +1,180 @@
...
+    const [refreshInterval, setRefreshInterval] = useState<number>(0);
```
mpetrunic: rename to refreshIntervalId

src/renderer/components/BalanceGraph/BalanceGraph.tsx
```diff
@@ -0,0 +1,180 @@
...
+        
+        const timeOnMount = new Date();
+        setCurrentHour(timeOnMount.getHours());
+        setCurrentDay(timeOnMount.getDay());
+        setCurrentMonth(timeOnMount.getMonth());
+
+        switch (interval) {
+            case IntervalEnum.HOUR: {
+                const localRefreshInterval = window.setInterval(awaitData, 60000);
+                setRefreshInterval(localRefreshInterval);
```
mpetrunic: extract those two lines to only be called once

src/renderer/components/BalanceGraph/BalanceGraph.tsx
```diff
@@ -0,0 +1,180 @@
...
+                        if(timeOnInterval.getHours()===0){
+                            setCurrentHour(timeOnInterval.getHours());
+                            awaitData();
+                        }
+                    } else {
+                        if(timeOnInterval.getHours()>currentHour){
+                            setCurrentHour(timeOnInterval.getHours());
+                            awaitData();
+                        }
+                    }
```
mpetrunic: You can make this code more generic by storing lastRefreshTime = new Date() in state
and on each interval do `const diffInSeconds = new Date().getTime()/ 1000 - lastRefreshTime.getTime()/ 1000`

and then do switch case:
```
switch(interval)
case Day:
if(diffInSeconds >= dayToSeconds) {
   refresh()
}
```

PR #92, opened 2019-12-09
Feature/redux connect w db

jogamod:
Screens tests are not working.


package.json
```diff
@@ -42,6 +42,7 @@
     "test:unit": "cross-env ELECTRON=1 yarn run jest --coverage \"test/unit\"",
     "test:e2e": "cross-env ELECTRON=1 yarn run jest \"test/e2e\"",
     "test:components": "yarn run jest --ci \"test/components\"",
+    "test:redux": "cross-env ELECTRON=1 yarn run jest \"test/redux\"",
```
mpetrunic: e2e?

src/main/database.ts
```diff
@@ -0,0 +1,31 @@
+import {CGDatabase} from "../renderer/services/db/api";
+// @ts-ignore
```
mpetrunic: ?

src/main/database.ts
```diff
@@ -0,0 +1,31 @@
...
+import {LevelDbController} from "../renderer/services/db/controller";
+import {config} from "@chainsafe/eth2.0-config/lib/presets/minimal";
+import {IService} from "../renderer/services/interfaces";
+import {CGAccount} from "../renderer/models/account";
+
+export class DatabaseHandler implements IService {
+    private database: CGDatabase;
+
+    public constructor() {
+        this.database = new CGDatabase({config, controller: new LevelDbController({name: "test-level.db"})});
```
mpetrunic: introduce config object as constructor param with default values.
I would suggest having:
- name - database name
- controller - so you can stub it out in tests

src/main/database.ts
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    public saveToDatabase(id: string, input: any): void {
+        this.database.account.set(id, input);
+    }
+
+    public async getFromDatabase(id: string): Promise<CGAccount | null> {
+        const account = await this.database.account.get(id);
+        return account;
+    }
```
mpetrunic: those two methods are not ok.
I would suggest something like 
```javascript
start() {
   await database.start();
   ipcMain.on(DATABASE_SAVE_EVENT, this.handleSaveEvent)
}

stop() {
   await database.stop();
   ipcMain.removeListener(DATABASE_SAVE_EVENT, this.handleSaveEvent)
}

private handleSaveEvent = (event, args: {table, id, serializedData}) => {
   this.database[table].save(id, serializedData);
}
```

Something along these lines. Might be good idea to leave CGDatabase and keep just DbController in main. In that case CGDatabase would handle serialization and just send stuff to database controller over ipc.

That said, ignore this for now, I'll open separate issue that we can deal with after we merge this.

src/renderer/actions/deposit.ts
```diff
@@ -0,0 +1,79 @@
...
+import {PrivateKey} from "@chainsafe/bls/lib/privateKey";
+import {EthersNotifier} from "../services/deposit/ethers";
+
+// Generate deposit action
+export const generateDepositAction = (networkConfig: INetworkConfig, amount: string) => {
+    return (dispatch: Dispatch<IGenerateDepositAction>, getState: () => IRootState): void => {
+        const signingKey = getState().register.signingKey;
+        const keyPair = new Keypair(PrivateKey.fromHexString(signingKey));
+        // Call deposit service and dispatch action
+        const depositData = generateDeposit(keyPair, Buffer.alloc(48, 1, "hex"), amount);
```
mpetrunic: second param should be withrawal public key from state

src/renderer/actions/deposit.ts
```diff
@@ -0,0 +1,79 @@
...
+        const depositTx = DepositTx.generateDepositTx(
+            depositData,
+            networkConfig.contract.address,
+            networkConfig.config,
+            amount);
+        const txData = typeof depositTx.data === "object" ?
+            `0x${depositTx.data.toString("hex")}` : `0x${depositTx.data}`;
+
+        dispatch(setDepositGenerated(true));
```
mpetrunic: isn't this reduntant since you can just check is there is txData in state?

src/renderer/actions/deposit.ts
```diff
@@ -0,0 +1,79 @@
...
+        dispatch(setDepositTransactionData(txData));
+    };
+};
+
+// Verify deposit action
+export const verifyDepositAction = (networkConfig: INetworkConfig, networkName: string) => {
```
mpetrunic: network name should be in networkConfig

src/renderer/actions/register.ts
```diff
@@ -1,14 +1,97 @@
...
+    };
+
+export const setWithdrawalKey = (withdrawalKey: string): IWithdarawalKeyAction => ({
+    type: RegisterActionTypes.STORE_WITHDRAWAL_KEY, payload: {withdrawalKey}
+});
+export interface IStoreWithdrawalKeyPayload {
+    withdrawalKey: string;
+}
+
+export interface IWithdarawalKeyAction extends Action<RegisterActionTypes> {
```
mpetrunic:
```suggestion
export interface IWithdrawalKeyAction extends Action<RegisterActionTypes> {
```

src/renderer/actions/register.ts
```diff
@@ -1,14 +1,97 @@
...
+    signingKey: string;
+}
+
+export interface ISigningKeyAction extends Action<RegisterActionTypes> {
+    payload: IStoreSigningKeyPayload;
+}
+
+// Withdrawal key action
+export const storeWithdrawalKeyAction = (withdrawalKey: string) =>
+    (dispatch: Dispatch<IWithdarawalKeyAction>): void => {
```
mpetrunic:
```suggestion
    (dispatch: Dispatch<IWithdrawalKeyAction>): void => {
```

src/renderer/constants/action-types.ts
```diff
@@ -1,3 +1,12 @@
 export enum RegisterActionTypes {
-    STORE_SIGNING_KEY_MNEMONIC = "ADD_MNEMONIC"
+    STORE_SIGNING_KEY_MNEMONIC = "ADD_MNEMONIC",
```
mpetrunic:
```suggestion
    STORE_SIGNING_KEY_MNEMONIC = "ADD_SIGNING_MNEMONIC",
```

src/renderer/constants/db.ts
```diff
@@ -0,0 +1,3 @@
+export const ACCOUNT_ID = "test";
```
mpetrunic: ?

src/renderer/constants/keystore.ts
```diff
@@ -0,0 +1 @@
+export const KEYSTORE_DEFAULT_DIRECTORY = "./keystore";
```
mpetrunic: should be user home directory, check some electron libs to determine best place across platforms

src/renderer/containers/Onboard/DepositTx/DepositTxContainer.tsx
```diff
@@ -17,77 +20,139 @@ const depositContracts = [
...
+    // TODO there is a flag in redux "isDepositVisible" so component should wait until flag is set to true
+    private handleVerify = (): void => {
+        const {selectedContractAddress, selectedNetworkIdx} = this.state;
+        // FIXME  pass network name from select but find names that ether.js supports
+        // You can use any standard network name
+        //  - "homestead"
+        //  - "rinkeby"
+        //  - "ropsten"
+        //  - "kovan"
+        //  - "goerli"
```
mpetrunic: We need eth1NetworkId in networkConfig and we need to have some utility function that maps that it to ethers network name

src/renderer/reducers/deposit.ts
```diff
@@ -0,0 +1,36 @@
...
+import {Action} from "redux";
+
+export interface IDepositState {
+    isDepositGenerated: boolean;
+    isDepositVisible: boolean;
+    txData: string;
+}
+
+const initialState: IDepositState = {
+    isDepositGenerated: false,
```
mpetrunic: I think it's much cleaner to expose this as function `if(txData) return true`

src/renderer/reducers/deposit.ts
```diff
@@ -0,0 +1,36 @@
...
+    isDepositVisible: false,
```
mpetrunic: Seems like it would be more useful to be amount deposited as <32 amounts will need special status in dashboard that they need more eth

src/renderer/reducers/register.ts
```diff
@@ -1,21 +1,37 @@
...
-    mnemonic: string
+    mnemonic: string,
+    signingKey: string,
+    withdrawalKey: string
 }
 
-const initalState: IRegisterState = {
-    mnemonic: ""
+const initialState: IRegisterState = {
+    mnemonic: "",
```
mpetrunic: Do we need this, you can always convert immediately to signingKey?

jogamod: Hmmm, @irubido  did this in the first place, I also was wondering if we gonna need it, but then I realized that he will maybe use it in mnemonic verify? Ivan can you confirm this

src/main/database.ts
jogamod: Should I have a name alongside controller (because controller already includes name)?

src/renderer/reducers/register.ts
mpetrunic: @jogamod yes, we need this for mnemonic verifying. I'll open an issue to pass mnemonic by url param

src/main/database.ts
mpetrunic: well you need  name here so you can pass it to controller. Not sure if I understood your question

src/main/database.ts
```diff
@@ -0,0 +1,51 @@
...
+    this.database = new CGDatabase({ config, controller: new LevelDbController({ name: this.databasePath }) });
+  }
+
+  public async start(): Promise<void> {
+    await this.database.start();
+    //ipcMain.on(DATABASE_SAVE_EVENT, this.handleSaveEvent)
+  }
+
+  public async stop(): Promise<void> {
+    deleteFolderRecursive(this.databasePath)
```
mpetrunic: move this logic to tests, we don't wan't to delete database in production

src/main/database.ts
```diff
@@ -0,0 +1,51 @@
...
+import { CGAccount } from "../renderer/models/account";
+import fs from 'fs'
+
+export class DatabaseHandler implements IService {
+  private database: CGDatabase;
+  private databasePath: string;
+
+  public constructor() {
+    const random = Math.floor(100000 + Math.random() * 900000)
+    this.databasePath = `test-level-${random}.db`
```
mpetrunic: you can create config object file where it maps env variables and provides default values.
Because we don't wan't random database names in production

mpetrunic: you can then pass name into setApp and set it there as env in each test

src/renderer/reducers/register.ts
irubido: yes, I used it to pass mnemonic

PR #94, opened 2019-12-10
Feature/display user validators

MakMuftic:
WIP - DashboardContainer is finished, but all data is hardcoded as an array. 
@mpetrunic I opened it just you can review the container itself.


src/renderer/containers/Dashboard/DashboardContainer.tsx
```diff
@@ -1,11 +1,125 @@
...
+            deposit: 30
+        },{
+            name: "V3",
+            status: "Not Working",
+            publicKey: "0x1d32a7822345564",
+            deposit: 30
+        }];
+    }
+
+    private getNetworks(): Array<string> {
```
mpetrunic: this will be just exported object with networkId -> name mapping

src/renderer/components/Dropdown/Dropdown.tsx
```diff
@@ -31,19 +36,31 @@ export const Dropdown: React.FunctionComponent<IDropdownProps> = (props: IDropdo
...
+                                // mapping object
+                                Object.values(props.options).map(value => {
+                                    return <div
+                                        key={value}
+                                        onClick={(): void => {props.onChange(getSelectedIndex(value));showHide();}}
+                                        className={
+                                            `dropdown-item
+                                            ${visible}
+                                            ${getSelectedIndex(value)===props.current?"selected":""}`}
+                                    >{value}</div>;})
```
mpetrunic: extract option rendering to separate method

src/renderer/components/Dropdown/Dropdown.tsx
```diff
@@ -31,19 +36,31 @@ export const Dropdown: React.FunctionComponent<IDropdownProps> = (props: IDropdo
...
-                            {props.options.map(option =>{
-                                return <div 
-                                    key={option} 
-                                    onClick={(): void => {props.onChange(getSelectedIndex(option));showHide();}} 
-                                    className={`dropdown-item 
-                                ${visible}
-                                ${getSelectedIndex(option)===props.current?"selected":""}`} 
-                                >{option}</div>;
-                            })}
+                            { Array.isArray(props.options) ?
```
mpetrunic: Wouldn't it be easier if you convert array to object and call onChange with key?
Than you don't need getSelectedIndex as key will always be either prop in object or index in array?

PR #112, opened 2019-12-13
Feature/export keystore

MakMuftic:


package.json
```diff
@@ -189,12 +189,13 @@
     "levelup": "4.3.2",
     "node-cmd": "^3.0.0",
     "prop-types": "^15.7.2",
+    "quickly-copy-file": "^1.0.0",
```
mpetrunic: really?
I don't think we need anothe library for such simple task

src/renderer/containers/Dashboard/DashboardContainer.tsx
```diff
@@ -1,11 +1,160 @@
...
+                                        onRemoveClick={(): void => {this.onRemoveValidator(index);}}
+                                        onExportClick={(): void => {this.onExportValidator(index);}}
+                                    />
+                                </div>;
+                            })}
+                    </div>
+                    <Notification
+                        isVisible={this.state.notification.visible}
+                        level={this.state.notification.level}
+                        title={this.state.notification.title}
```
mpetrunic: If you wan't to display second notification it will just rename current?

src/renderer/containers/Dashboard/export.ts
```diff
@@ -0,0 +1,39 @@
...
+            status: {
+                level: copyResult.success ? Level.INFO : Level.ERROR,
+                message: copyResult.success ?
+                    `Successfully exported keystore for validator ${validator.name} to ${savePath}.` :
+                    `Export failed: ${copyResult.message}.`
+            }
+        };
+    }
+    // destination path not selected
+    return {};
```
mpetrunic: it's probably better to return null

src/renderer/services/utils/file-utils.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export interface ICopyFileStatus {
+    success: boolean;
+    message: string
+}
+
+export function copyFile(fromPath: string, toPath: string, readOptions = "utf-8"): ICopyFileStatus {
+    try {
+        const fileContent = fs.readFileSync(fromPath, readOptions);
+        fs.writeFileSync(toPath, fileContent);
```
mpetrunic: We ar enot int 2003, use pipes or native copy:
https://stackoverflow.com/questions/11293857/fastest-way-to-copy-file-in-node-js

:smile:

PR #114, opened 2019-12-16
Add rest api json conversion

### LINKED ISSUE #39: Fix rest client
There is new package: https://www.npmjs.com/package/@chainsafe/eth2.0-utils
It has method to convert from rest json to eth2-types:
https://github.com/ChainSafe/lodestar/blob/master/packages/eth2.0-utils/src/converters/json.ts#L46

Additionally:
- update unit tests to reflect that
- update rest of lodestar packages to latest version

mpetrunic:
resolves #39


src/renderer/services/api/eth2.ts
```diff
@@ -41,27 +44,39 @@ export class Eth2 implements IBeaconAPIClient {
     }
 
     public async fetchGenesisTime(): Promise<uint64> {
-        return this.httpClient.get<uint64>(FETCH_GENESIS_TIME);
+        return new BN(await this.httpClient.get<string>(FETCH_GENESIS_TIME));
```
morrigan: isn't this BN return type?

mpetrunic: api will return string as it's only format that can contain big numbers, uin64 = BN

PR #116, opened 2019-12-18
Password dialog

### LINKED ISSUE #98: Password dialog
- expose method which will open native dialog(checkout electron prompts) which is suppose to block main window
- prompt (dialog) should have submit and cancel button
- password input should be hidden
- method should be async and return password or null
- submit is suppose to check password before returning and display error message if invalid

MakMuftic:
resolves #98


src/renderer/components/Prompt/InputPrompt.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+    const [visible, setVisible]=useState(false);
+
+    const handler = () => setVisible(true);
+    button("show prompt", handler);
+
+    return <div>
+        <InputPrompt
+            display={visible}
+            onSubmit={(): ISubmitStatus => { return {valid: true};}}
+            onCancel={() => {}}
```
mpetrunic: setVisible(false)

src/renderer/components/Input/InputForm.tsx
```diff
@@ -25,14 +26,16 @@ export const InputForm: React.FunctionComponent<IInputFormProps> = (props: IInpu
     return( 
         <form>
             <div className="label">{props.label}</div>
-            <input 
+            <input
```
mpetrunic: Can you disable form submitting when pressing "Enter"

mpetrunic: better yet it should call some onSubmit function :smile:

src/renderer/components/Prompt/PasswordPrompt.tsx
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export const PasswordPrompt: React.FunctionComponent<IPasswordPromptProps> = (props: IPasswordPromptProps) => {
+    return(
+        <InputPrompt
+            onSubmit={props.onSubmit}
+            onCancel={props.onCancel}
+            display={props.display}
+            title={"Confirm password"}
+            placeholder={"Please reenter your password..."}
```
mpetrunic: enter*

PR #117, opened 2019-12-19
Database over IPC

### LINKED ISSUE #97: Database over IPC
Renderer:
- implement new DatabaseController (IpcDatabaseController) which passes all data to appropriate ipc channels
- use replace LevelDatabaseController with IpcDatabaseController

Main:
- move LevelDatabaseController from renderer
- Create LevelDatabaseController wrapper which listens on ipc events and passes given data to LevelDatabaseController

Overview of db arch would be:
renderer process -> DatabaseApi -> IpcDatabaseController -> ipc event -> main process -> ipcListener -> LevelDatabaseController

### LINKED ISSUE #110: Complete registration
- fix database not open error on password confirmation
- remove consent step from register navigation

mpetrunic:
- database is controlled from main process
- renderer communicates with database over ipc
- upgrade electron to version 7
- fix keystore location and registration
- add redux logger

resolves #97 
resolves #110


src/main/db/controller/impl/level.ts
```diff
@@ -86,4 +96,13 @@ export class LevelDbController extends EventEmitter implements IDatabaseControll
                 });
         });
     }
+
+    private getDatabaseLocation(): string {
+        try {
+            fs.mkdirSync(this.opts.location, {recursive: true});
+        } catch (e) {
+            warn("creating database directories", e);
```
morrigan: Shouldn't this be an error ?

src/renderer/actions/register.ts
```diff
@@ -65,32 +65,40 @@ export interface IWithdrawalKeyAction extends Action<RegisterActionTypes> {
...
         // 1. Save to keystore
-        const signingKey = getState().register.signingKey;
-        V4Keystore.create(
-            `${KEYSTORE_DEFAULT_DIRECTORY}/${getV4Filename()}.json`, 
-            password, new Keypair(PrivateKey.fromHexString(signingKey))
+        dispatch(startRegistrationSubmission());
+        const signingKey = PrivateKey.fromBytes(
+            Buffer.from(getState().register.signingKey.slice(2), "hex")
+        );
+        const accountDirectory = path.join(getConfig().storage.accountsDir, "default");
```
morrigan: maybe we should export account handling (location and keystore management) somewhere else? I guess we will probably use it for login and unlocking.

morrigan: also not using DEFAULT_ACCOUNT var

src/renderer/actions/register.ts
```diff
@@ -65,32 +65,40 @@ export interface IWithdrawalKeyAction extends Action<RegisterActionTypes> {
...
+        await V4Keystore.create(
+            path.join(accountDirectory, PublicKey.fromPrivateKey(signingKey).toHexString() + ".json"),
+            password, new Keypair(signingKey)
         );
 
         // 2. Save account to db
         const account = new CGAccount({
-            name: "Test Account",
-            directory: KEYSTORE_DEFAULT_DIRECTORY,
+            name: "Default",
```
morrigan: also DEFAULT_ACCOUNT ?

src/main/db/controller/impl/level.ts
mpetrunic: This will throw if database already exists

src/renderer/actions/register.ts
mpetrunic: > maybe we should export account handling (location and keystore management) somewhere else? I guess we will probably use it for login and unlocking.

Hmm not sure where else you would use account and keystore creation though. I guess we can see later on

src/renderer/actions/register.ts
mpetrunic: We don't use this field, it will be used later on if we introduce multi-account stuff

mpetrunic:
> Commented on a few things that I've noticed. How did this work before without IPC? Or this is the first time to connect renderer to db?

It had some hardcoded ipc call to amke it work, but it didn't work anyways

PR #118, opened 2019-12-19
Tooltip fix, Input UI fix

irubido:
Tooltip:

	ReactTooltip moved to CopyButton component

	added New prop clicked?: boolean;  to <CopyButton> and <CopyField>/<MnemonicCopyField>

	handleCopy method in Signing method to handle state of copied mnemonic and clipboard

InputForm UI fix: 

	width set to 100%

	text doesn't overlap with validation checkmark

Login Screen UI fix:

	margin-top from “OR” fixed to match with design in Sigma


src/renderer/components/Button/ButtonAction.tsx
```diff
@@ -2,11 +2,16 @@ import React from "react";
...
 import add from "../../assets/img/buttons/Add.svg";
 import copyDefault from "../../assets/img/buttons/CopyDefault.svg";
+import ReactTooltip from "react-tooltip";
 
 export interface IActionButtonProps {
     onClick?: () => void;
 }
 
+export interface ICopyButtonProps extends IActionButtonProps {
+    clicked?: boolean;
```
mpetrunic: I don't think there is need for somebody else to set this. It would be better if that component uses state internally and resets after 1s to "Copy All" text

PR #121, opened 2019-12-19
Successful mnemonic verification

### LINKED ISSUE #105: Nothing happens after successful signing key mnemonic verification
Should happen:
   - save private key in redux state
   - navigate to withrawal key flow
   - remove verification route from history (back button should lead to generate mnemonic)

morrigan:
Resolves #105 
with created Withdrawal key initial step.


morrigan:
@mpetrunic I've realized that I'm missing paragraph that says `If you skip, we won’t be able to generate a deposit transaction.` meaning that the skip button should go to create password screen?

mpetrunic:
@morrigan yes

src/renderer/components/Button/ButtonStandard.tsx
```diff
@@ -6,6 +6,7 @@ export interface IButtonProps {
     onClick?: () => void;
     large?: boolean;
     buttonId?: string;
+    dataField?: any;
```
mpetrunic: don't use any type. You could do something like `string|number`

mpetrunic:
```suggestion
    dataField?: string | number;
```

morrigan:
@mpetrunic thanks for the fix! :dancer:

PR #134, opened 2020-01-14
Read validators from logged in CGAccount

irubido:
- Dashboard now shows validators from logged in CGAccount.

- Dashboard changed from class component to functional component

- PrivateKeyField now compares input password with password from CGAccount method isCorrectPassword()

problems: 

storybook has problems with node 'fs', -> storybooks that await password check from db are commented, Validator and PrivateKeyField

Validator.spec.ts fails on ValidatorSimple -> commented
Test suite failed to run
            .TypeError: Cannot read property 'randomBytes' of undefined
             > 5 | import {Keystore, IKeystore} from "@nodefactory/bls-keystore";

- To show Validators from CGAccount password is needed for unlock method, so I saved password to redux.
Is that intended flow or should unlock method persist to other screens until end of session?

- Test error:
Ran all test suites matching /test\/components/i.
error Command failed with exit code 1.


src/renderer/containers/Dashboard/DashboardContainer.tsx
```diff
@@ -27,144 +29,135 @@ export interface IValidator {
...
+        const validatorArray = [];
+
+        const validatorsData = await database.account.get(DEFAULT_ACCOUNT);
+        if(validatorsData){
+            await validatorsData.unlock(props.auth.password); /** Password from redux */
+            const x =validatorsData.getValidators();
+            
+            for (let i = 0; i < x.length; i++) {
+                validatorArray.push({
+                    name: "TODO name",
```
mpetrunic: CGAccount should have name property

src/renderer/containers/Dashboard/DashboardContainer.tsx
```diff
@@ -27,144 +29,135 @@ export interface IValidator {
...
+
+    const getValidators = async (): Promise<void> => {
+        const validatorArray = [];
+
+        const validatorsData = await database.account.get(DEFAULT_ACCOUNT);
+        if(validatorsData){
+            await validatorsData.unlock(props.auth.password); /** Password from redux */
+            const x =validatorsData.getValidators();
+            
+            for (let i = 0; i < x.length; i++) {
```
mpetrunic: use map  instead of for loop

src/renderer/containers/Dashboard/DashboardContainer.tsx
```diff
@@ -27,144 +29,135 @@ export interface IValidator {
...
+            });
+        }
+    };
+
+    const getValidators = async (): Promise<void> => {
+        const validatorArray = [];
+
+        const validatorsData = await database.account.get(DEFAULT_ACCOUNT);
+        if(validatorsData){
+            await validatorsData.unlock(props.auth.password); /** Password from redux */
```
mpetrunic: validatorData  = this.props.auth;

src/renderer/actions/auth.ts
```diff
@@ -15,4 +15,18 @@ export interface IStoreAuthPayload {
 }
 export interface IStoreAuthAction extends Action<AuthActionTypes> {
     payload: IStoreAuthPayload;
+}
+//Storing password
+export const storePasswordAction = (password: string) =>
```
mpetrunic: you don't need this, unlocked account should be in state

mpetrunic:
you have some CI errors though

PR #139, opened 2020-01-29
Prevent accidental window closing

irubido:
- default dialog when exiting app
- TODO / needs validator status in redux as condition


src/main/gui/window.ts
```diff
@@ -53,6 +54,22 @@ export async function createWindow(): Promise<void> {
         });
     }
 
+    win.on("close", (e: Electron.Event) => {
+        // TODO / Validator status - check if there is validator with status ACTIVE/VALIDATING
+        if (win!==null){
```
mpetrunic: && NODE_ENV === "production"

morrigan:
@mpetrunic I am not sure what's wrong with this test but worked before merging master

mpetrunic:
@morrigan not sure, does it work locally with merged master?

PR #143, opened 2020-02-03
Global notification system

irubido:
- added NotificationRenderer container

- NotificationStacked is a temporary copy of Notification component, uses slightly different css to adjust for Notification stack in bottom right corner
new issue -> replace all Notifications with redux Notifications

- redux state: 
added notificationArray object, containing 2 arrays: "stacked" and "other"
if notification has bottom right position props it will be added to stacked array that shows notification stacked bottom up
notificationArray.other contains all other notifications on different positions

- redux actions:
addNotificationAction - adds notification to state, and removes it after a set amount of time(if provided in props)
removeNotificationAction - removes notification from state, onClick() or by expireTime if provided in addNotificationAction


src/renderer/NotificationRenderer.tsx
```diff
@@ -0,0 +1,69 @@
...
+                <div className={"notification-stacked-container right bottom"}>
+                    {arrays.stacked.map((n, index) =>
+                        <NotificationStacked
+                            key={index}
+                            title={n.title}
+                            isVisible={n.isVisible}
+                            level={n.level}
+                            horizontalPosition={n.horizontalPosition}
+                            verticalPosition={n.verticalPosition}
+                            onClose={(): void => {props.removeNotification(index, true);}}
```
mpetrunic: It's probably safer, if each notification has some id(maybe id should include some container identification) instead of dynamic index

src/renderer/actions/notification.ts
```diff
@@ -0,0 +1,53 @@
...
+import {INotificationState} from "../reducers/notification";
+import {Horizontal, Vertical} from "../components/Notification/NotificationEnums";
+
+//Add Notification
+/** expireTime is in ms */
+export const storeNotificationAction = (notification: INotificationState, expireTime?: number
+) =>
+    (dispatch: Dispatch<IStoreNotificationAction>): void => {
+        let isStacked = false;
+        if(notification.horizontalPosition===Horizontal.RIGHT
```
mpetrunic: add comment that only bottom right is stacked

src/renderer/actions/notification.ts
```diff
@@ -0,0 +1,53 @@
...
+export const storeNotificationAction = (notification: INotificationState, expireTime?: number
+) =>
+    (dispatch: Dispatch<IStoreNotificationAction>): void => {
+        let isStacked = false;
+        if(notification.horizontalPosition===Horizontal.RIGHT 
+            && notification.verticalPosition===Vertical.BOTTOM)
+            isStacked = true;
+        dispatch(setNotification( notification
+        ));
+        expireTime ? setTimeout(dispatch, expireTime, setRemoveNotification(0,isStacked)) : null;
```
mpetrunic: why are you always removing notification with index 0?

src/renderer/actions/notification.ts
```diff
@@ -0,0 +1,53 @@
...
+    stackedArray: boolean
+): IRemoveNotificationAction => ({
+    type: NotificationActionTypes.REMOVE_NOTIFICATION, payload: {
+        index,
+        stackedArray
+    }
+});
+export interface IRemoveNotificationPayload {
+    index: number,
+    stackedArray: boolean
```
mpetrunic: this should be unnessesary with notification id

src/renderer/reducers/notification.ts
```diff
@@ -0,0 +1,66 @@
...
+    other: []
+};
+
+export const notificationReducer = (
+    state = initialState, 
+    action: Action<NotificationActionTypes>): INotificationStateObject => {
+
+    // Copy of current notification state arrays
+    const newStackedArray = state.stacked.slice();
+    const newOtherArray = state.other.slice();
```
mpetrunic: This will always refresh UI generated by both arrays, you should create new array only if you are going to add something

src/renderer/reducers/notification.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    switch (action.type) {
+        case NotificationActionTypes.ADD_NOTIFICATION: {
+            const actionProps = (action as IStoreNotificationAction).payload;
+
+            if(actionProps.horizontalPosition===Horizontal.RIGHT && actionProps.verticalPosition===Vertical.BOTTOM) {
```
mpetrunic: you have this condition is bunch of places, extract it to method isStackedNotification position

src/renderer/actions/notification.ts
irubido: Mistake, I was testing with buttons dispatching same action with same expireTime, so it worked perfectly only for that case, because latest notification was always with index 0. 

I'll fix it so it removes notification based on ID.

PR #144, opened 2020-02-03
Validation text fix

irubido:


src/renderer/containers/Onboard/SigningKey/Import/validation.ts
```diff
@@ -5,7 +5,7 @@ import {StringSchema} from "@hapi/joi";
 
 export const PRIVATE_KEY_NOT_STRING_MESSAGE="Value must be non empty string";
 export const PRIVATE_KEY_WRONG_CHARACTERS_MESSAGE = "Private key must contain hex characters only";
-export const PRIVATE_KEY_WRONG_LENGTH_MESSAGE = `Private key should have ${SECRET_KEY_LENGTH} bytes`;
+export const PRIVATE_KEY_WRONG_LENGTH_MESSAGE = `Private key should have ${SECRET_KEY_LENGTH} characters`;
```
mpetrunic: SECRET_KEY_LENGTH*2 characters

each byte is represented as 2 hex characters

mpetrunic:
I think we have same error for public key

src/renderer/containers/Onboard/SigningKey/Import/validation.ts
irubido: SECRET_KEY_LENGTH constant was set to 64, so i only changed text

"export declare const SECRET_KEY_LENGTH = 64;"

I'll change constant to 32 then and set {SECRET_KEY_LENGTH*2} characters

mpetrunic: Hmm that's odd, can you check where else this constant is used?

If nowhere, set it to 32 and multiply by two here

irubido: I changed it and committed changes, but it does not show on pull request, strange, and I couldn't see it git either. I checked, its only used for this case.

mpetrunic: I think we should still set to 32 as this is missleading

PR #145, opened 2020-02-04
Fix, Delete validators

irubido:
now working as intended, after deletion, change is dispatched to CGAccount in app state

### DIFF
--- a/src/renderer/containers/Dashboard/DashboardContainer.tsx
+++ b/src/renderer/containers/Dashboard/DashboardContainer.tsx
@@ -14,6 +14,8 @@ import {Routes, OnBoardingRoutes} from "../../constants/routes";
 import {ConfirmModal} from "../../components/ConfirmModal/ConfirmModal";
 import {V4Keystore} from "../../services/keystore";
 import * as path from "path";
+import {storeAuthAction} from "../../actions/auth";
+import {bindActionCreators, Dispatch} from "redux";
 
 type IOwnProps = Pick<RouteComponentProps, "history" | "location">;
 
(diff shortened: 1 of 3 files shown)


src/renderer/containers/Dashboard/DashboardContainer.tsx
```diff
@@ -64,16 +66,18 @@ const Dashboard: React.FunctionComponent<IOwnProps & Pick<IRootState, "auth">> =
...
             const selectedValidatorPublicKey = validators[selectedValidatorIndex].publicKey.toHexString();
             const selectedV4Keystore = new V4Keystore(
-                path.join(validatorsData.directory,selectedValidatorPublicKey, ".json"));
+                path.join(validatorsData.directory,selectedValidatorPublicKey + ".json"));
             selectedV4Keystore.destroy();
+            props.auth.auth.removeValidator(selectedValidatorIndex);
+            props.storeAuth(props.auth.auth);
         }
-        setValidators(validators.splice(selectedValidatorIndex, 1));
+        getValidators();
```
mpetrunic: rename to loadValidators since you aren't getting anything

src/renderer/models/account.ts
```diff
@@ -112,6 +112,9 @@ export class CGAccount implements IAccount {
     public addValidator(validator: Keypair): void {
         this.validators.push(validator);
     }
+    public removeValidator(index: number): void {
+        this.validators.splice(index, 1);
```
mpetrunic: But if you login again you will have that validator again?

I think we need to delete physical keystore as well

irubido: It deletes physical keystore in Dashboard with .destroy() method.
I made this remove validator method just to remove it from CGAccount saved in redux.

`const selectedV4Keystore = new V4Keystore(
                path.join(validatorsData.directory,selectedValidatorPublicKey + ".json"));
            selectedV4Keystore.destroy();
            props.auth.auth.removeValidator(selectedValidatorIndex);
            props.storeAuth(props.auth.auth);`

morrigan:
@mpetrunic approvy again pls

PR #152, opened 2020-02-07
Connect to beacon chain

### LINKED ISSUE #148: Support connecting to network
- implement container.ts so it can run prysmatic beacon node in docker
   - should provide options to connect to various networks (just public for now)
- ability to stream logs
- add redux functions that should run implemented container in background (another thread?!)
- add redux functions to stop and restart that container

morrigan:
Resolves #148

### DIFF
--- a/src/renderer/services/docker/chain.ts
+++ b/src/renderer/services/docker/chain.ts
@@ -1,0 +1,11 @@
+import { Container } from './container';
+
+export class ValidatorNetwork extends Container {
+    public static connectToPrysmNetwork() {
+        return new ValidatorNetwork({
+            image: "gcr.io/prysmaticlabs/prysm/validator:latest",
+            name: "PrysmValidator",
+            restart: "unless-stopped"
+        });
+    }
+}
--- a/src/renderer/services/docker/chain.ts
+++ b/src/renderer/services/docker/chain.ts
@@ -1,6 +1,6 @@
-import { Container } from './container';
+import {Container} from "./container";
 
-type LogType = 'info' | 'error'
+type LogType = "info" | "error";
 type LogCallbackFunc = (type: LogType, message: string) => void;
 
 export enum SupportedNetworks {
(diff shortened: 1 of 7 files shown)


test/e2e/services/docker/chain.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    let beaconChain: BeaconChain;
+
+    // clean up created container
+    afterEach(async () => {
+        if (await BeaconChain.isDockerInstalled()) {
+            const isRunning = await beaconChain.isRunning();
+            if (isRunning) {
+                await beaconChain.stop();
+            }
+            await runCmdAsync("docker rm Prysm-beacon-node").catch();
```
mpetrunic: use SupportedNetwork.Prysm so this doesn't break if we change

test/e2e/services/docker/chain.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+            await runCmdAsync("docker volume rm Prysm-chain-data").catch();
```
mpetrunic: how does await and .catch() works?

src/renderer/services/docker/chain.ts
```diff
@@ -0,0 +1,42 @@
...
+
+type LogType = "info" | "error";
+type LogCallbackFunc = (type: LogType, message: string) => void;
+
+export enum SupportedNetworks {
+    PRYSM = "Prysm",
+}
+
+export class BeaconChain extends Container {
+    public static instance: BeaconChain;
```
mpetrunic: Hmm i think better solution is to have singleton DockerContainers where each "Container" class could register itself.
In that way you could generate multiple BeaconChain containers like for local and public testnet at the same time

src/renderer/services/docker/chain.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    public static async startPrysmBeaconChain(waitUntilReady = false): Promise<BeaconChain> {
+        const bc = new BeaconChain({
+            image: "gcr.io/prysmaticlabs/prysm/beacon-chain:latest",
+            name: `${SupportedNetworks.PRYSM}-beacon-node`,
+            restart: "unless-stopped",
+            ports: ["4000:4000", "13000:13000"],
```
mpetrunic: I don't think port 13000 needs to be exposed to host.
Also user should be able to configure on which port does he wan't port 4000 to be exposed(aka this should be configurable by user)

morrigan:
@mpetrunic pls

PR #165, opened 2020-02-13
Network selection in onboarding flow

### LINKED ISSUE #154: Connecting to beacon node network in validator creation
Before deposit, user should select network and that should be stored in Redux.

After deposit, user should select if running own beacon node or connecting to existing one (design exists)

Run your own node option should start beacon chain docker based on the previous network selection with customization of local ports.

morrigan:
Resolves #154

- Moved configure screen with beacon chain settings before Deposit
- Added network dropdown in configure screen, made dropdown disabled in Deposit
- Start beacon chain on selecting to run own node with desired settings


src/renderer/reducers/network.ts
```diff
@@ -0,0 +1,22 @@
...
+    state = initialState,
+    action: ISaveSelectedNetworkAction): INetworkState => {
+
+    switch (action.type) {
+        case NetworkActionTypes.SELECT_NETWORK:
+            return {...state, selected: action.payload};
+        default:
+            return state;
+    }
+};
```
mpetrunic: In my opinion this should go to "register" state as global network selected would be one on dashboard

test/e2e/redux/reducers/register.reducer.spec.ts
```diff
@@ -22,6 +22,7 @@ const initalState: IRegisterState = {
     withdrawalMnemonic: "",
     withdrawalVerification: false,
     withdrawalKey: "",
+    network: "",
```
mpetrunic: maybe prysm as default value?

morrigan: It's default in the dropdown and will set it if nothing is selected. Idk, don't want to hardcode on many places 🤷‍♀

mpetrunic: It would make much more sense to set default here instead of that specific container

test/e2e/redux/reducers/register.reducer.spec.ts
```diff
@@ -22,7 +22,7 @@ const initalState: IRegisterState = {
     withdrawalMnemonic: "",
     withdrawalVerification: false,
     withdrawalKey: "",
-    network: "",
+    network: "Prysm",
```
mpetrunic: enum?

PR #166, opened 2020-02-14
Add deposit checking

### LINKED ISSUE #108: Deposit verification
- display loading overlay with button to cancel

mpetrunic:
- renamed some redux actions and states
- fixed loading screen zindex
- add checks and redirects for deposits
resolves #108


src/renderer/services/deposit/ethers.ts
```diff
@@ -49,30 +50,34 @@ export class EthersNotifier {
...
-        });
-    
-        return amountSum.cmp(etherToGwei(this.networkConfig.contract.depositAmount)) === 1;
+        try {
+            const filter = {
+                fromBlock: this.networkConfig.contract.deployedAtBlock,
+                address: this.networkConfig.contract.address
+            };
+            const logs = await this.provider.getLogs(filter);
+            const amountSum = new BN(0);
```
morrigan: We are using BigInt now :P

mpetrunic: not yet? :open_mouth:

PR #168, opened 2020-02-18
Update lodestar libs

### LINKED ISSUE #156: Update lodestar libraries
Libraries to update to latest version:
- @chainsafe/bls
- @chainsafe/eth2.0-config
- @chainsafe/eth2.0-types
- @chainsafe/eth2.0-utils
- @chainsafe/ssz

morrigan:
Resolves #156 

Also replaced BN.js with BigInt


.eslintrc.js
```diff
@@ -6,6 +6,9 @@ module.exports = {
         "node": true,
         "es6": true
     },
+    "globals": {
+        BigInt: true,
+    },
```
mpetrunic: I think this will provide better typehinting:
```
    "lib": [
      "esnext.bigint"
    ],
```

package.json
```diff
@@ -171,18 +171,18 @@
...
+    "@chainsafe/eth2.0-config": "^0.2.1",
+    "@chainsafe/eth2.0-types": "^0.2.1",
+    "@chainsafe/eth2.0-utils": "^0.1.1",
+    "@chainsafe/ssz": "^0.5.4",
     "@hapi/joi": "^16.1.7",
     "@nodefactory/bls-keystore": "^1.0.0-beta7",
     "axios": "0.19.0",
     "bech32": "1.1.3",
+    "bigint-buffer": "^1.1.5",
     "bip39": "3.0.2",
```
mpetrunic: I don't think we need this anymore as well as bech32

src/main/main.ts
```diff
@@ -12,6 +14,7 @@ app.on("ready", async function() {
     await Promise.all([
         db.start(),
         createWindow(),
+        initBLS(),
```
mpetrunic: Can you await initBLS before dba dn window start?

src/renderer/services/api/eth2.ts
```diff
@@ -83,12 +81,12 @@ export class Eth2 implements IBeaconAPIClient {
         validatorPubkey: BLSPubkey,
         pocBit: uint8,
         slot: Slot,
-        shard: Shard
+        shard: number,
```
mpetrunic: lel, leave this, I will correct this in Pr with validator api as this has become CommitteeIndex

test/unit/HDWallet.spec.ts
```diff
@@ -7,6 +8,10 @@ const publicKeyStr =
     "0x95c259ecefa8e60edb099a6568887dd9ba35afa75062d01a12fcc5a40d402eca57afcdc1cff7e34fbcd84dfea0848bb9";
 
 describe(".generate()", () => {
+    beforeAll(async () => {
+        await initBLS();
+    });
```
mpetrunic: Add this to setupFile:
https://jestjs.io/docs/en/configuration.html#setupfiles-array

.eslintrc.js
morrigan: I think you mean that for the tsconfig which I already did, this is for eslint

test/unit/HDWallet.spec.ts
morrigan: It doesn't work with async code, there are some hacks that don't look nice

src/renderer/models/account.ts
```diff
@@ -98,6 +98,7 @@ export class CGAccount implements IAccount {
             try {
                 return await keystore.decrypt(password);
             } catch (e) {
+                console.error(e);
```
mpetrunic: remove this :)

morrigan: but I had a hard time catching a bug because of it :D

mpetrunic: import and use electron-log :stuck_out_tongue:

morrigan: ah, didn't notice that one 👼

PR #169, opened 2020-02-18
Fix copy button memory leak and tooltip

### LINKED ISSUE #163: Copy Button - memory leak
- Can't perform a React state update on an unmounted component.

mpetrunic:
resolves #163


src/renderer/components/Button/ButtonAction.tsx
```diff
@@ -13,48 +14,62 @@ export interface ICopyButtonProps {
...
 
-    return(<button
-        className={"copy-button"} 
+    useEffect(() => {
+        return function cleanup(): void {
+            if (timeoutId) {
+                window.clearTimeout(timeoutId);
+            }
+        };
+    });
```
morrigan: You should memoize `timeoutId`. Also, I don't think you need to return a function...

PR #173, opened 2020-02-24
Connect BN validator

### LINKED ISSUE #159: Connect beacon node with validator
- db should be extended to hold BN details
- each validator should have attached BNs
- attached BNs needs to be displayed on dashboard

morrigan:
Resolves #159

### DIFF
--- a/src/renderer/actions/network.ts
+++ b/src/renderer/actions/network.ts
@@ -48,7 +48,7 @@ export const saveSelectedNetworkAction = (network: string): ISaveSelectedNetwork
     payload: network,
 });
 
-export const saveBeaconNode = (url: string) => {
+export const saveBeaconNodeAction = (url: string) => {
     return async (dispatch: Dispatch<Action<unknown>>, getState: () => IRootState): Promise<void> => {
         const signingKey = PrivateKey.fromBytes(
             Buffer.from(getState().register.signingKey.replace("0x",""), "hex")
--- a/src/renderer/actions/network.ts
+++ b/src/renderer/actions/network.ts
@@ -10,13 +10,16 @@ import database from "../services/db/api/database";
 
 export const startBeaconChainAction = (network: string, ports?: string[]) => {
     return async (): Promise<void> => {
+        let bc: BeaconChain;
         switch(network) {
             case SupportedNetworks.PRYSM:
-                await BeaconChain.startPrysmBeaconChain(ports);
+                bc = await BeaconChain.startPrysmBeaconChain(ports);
                 break;
             default:
-                await BeaconChain.startPrysmBeaconChain(ports);
+                bc = await BeaconChain.startPrysmBeaconChain(ports);
         }
+        const name = bc.getName();
+        await saveBeaconNodeAction("localhost", name);
     };
 };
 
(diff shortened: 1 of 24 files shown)


src/renderer/actions/network.ts
```diff
@@ -40,3 +46,41 @@ export const saveSelectedNetworkAction = (network: string): ISaveSelectedNetwork
...
+        const validatorAddress = signingKey.toPublicKey().toHexString();
+        await database.beaconNodes.set(
+            validatorAddress,
+            beaconNode,
+        );
+    };
+};
+
+export const loadedBeaconNode = (node: BeaconNode | null, validatorAddress: string): ILoadBeaconNodeAction => ({
+    type: NetworkActionTypes.LOAD_BEACON_NODE,
```
mpetrunic: Add plural everywhere. Like `load_beacon_nodes`

mpetrunic: I would also suggest to name this loadValidatorBeaconNodes since loadBeaconNodes will be used to fetch all available beacon nodes

src/renderer/models/beaconNode.ts
```diff
@@ -0,0 +1,11 @@
+export interface IBeaconNode {
+    url: string;
+}
+
+export class BeaconNode implements IBeaconNode {
+    public url: string;
```
mpetrunic: I think this should also hold info if it's local node like docker id or something (think about initializing dockers on startup and getting right docker from registry)

morrigan: Right, good input!

src/renderer/actions/network.ts
```diff
@@ -40,3 +46,18 @@ export const saveSelectedNetworkAction = (network: string): ISaveSelectedNetwork
     type: NetworkActionTypes.SELECT_NETWORK,
     payload: network,
 });
+
+export const saveBeaconNodeAction = (url: string, network?: string) => {
+    return async (dispatch: Dispatch<Action<unknown>>, getState: () => IRootState): Promise<void> => {
+        const localDockerName = network ? BeaconChain.getContainerName(network) : undefined;
```
mpetrunic: maybe use null?
What do we do if it's undefined/null?

src/renderer/components/Validator/ValidatorSimple.tsx
```diff
@@ -12,30 +14,56 @@ export interface IValidatorSimpleProps {
...
 
+    const renderBeaconNodes = (): React.ReactElement => (
+        <div className="validator-nodes">
+            <div className="node-container">
+                <div className="node-grid-container" >
+                    {props.nodes.map(node => (
+                        <NodeCard
+                            key={node.url}
+                            onClick={(): void=>{}}
+                            title="Beacon node"
```
mpetrunic: I was thinking, maybe we can put "<container name> docker container" and "Remote beacon node" depending on what type it is?

src/renderer/actions/network.ts
morrigan: It's an empty string if it is a remote beacon node. I put network here because of the container name :/

mpetrunic: What if it's a local node but container is down or offline?

morrigan: This is saved after starting the docker container. We don't have resuming container work anywhere for now

PR #177, opened 2020-03-13
Fix add validator flow

### LINKED ISSUE #155: Add new validator contains only signing key
Current flow:

1. ADd new validator
2. Import/Generate signing key
3. Enter password
4. Modal closed

Should go to withdrawal key and deposit...

morrigan:
Resolves #155 

Now adding new validator button leads through whole onboarding process, checking password (to create new keystores) and redirecting to dashboard. 

I have added new property in CGDatabase which is number of validators. The reason for this is that I can't trigger `useEffect` for loading validators and rerender with new validators list because `validators` is private. I am open to other ideas... Maybe the best would be not to store CGAccount as model in Redux store.


src/renderer/actions/register.ts
```diff
@@ -103,19 +104,26 @@ export interface IWithdrawalKeyAction extends Action<RegisterActionTypes> {
     payload: IStoreWithdrawalKeyPayload;
 }
 
+const saveKeystore = async(state: IRootState, password: string): Promise<string> => {
+    const signingKey = PrivateKey.fromBytes(
+        Buffer.from(state.register.signingKey.replace("0x",""), "hex")
```
mpetrunic: I think there is method stringToBuffer or something like that in our codebase

test/e2e/redux/reducers/register.reducer.spec.ts
```diff
@@ -24,6 +24,7 @@ const initalState: IRegisterState = {
     withdrawalVerification: false,
     withdrawalKey: "",
     network: SupportedNetworks.PRYSM,
+    addingNewValidator: false,
```
mpetrunic: i would prefer type: [REGISTER, ADD] instead of flags

src/renderer/containers/Dashboard/DashboardContainer.tsx
```diff
@@ -127,7 +116,7 @@ const Dashboard: React.FunctionComponent<IOwnProps & IInjectedProps & Pick<IRoot
         }
 
         loadValidators();
-    },[]);
+    },[props.auth.account && props.auth.account.numberOfValidators]);
```
mpetrunic: Cant you do here just `props.auth.account && props.auth.account.validators.length`?

src/renderer/models/account.ts
```diff
@@ -118,13 +120,16 @@ export class CGAccount implements IAccount {
                 this.validators.push(validator);
             }
         }
+        this.numberOfValidators = this.validators.length;
     }
 
     public addValidator(validator: Keypair): void {
         this.validators.push(validator);
+        this.numberOfValidators++;
```
mpetrunic: I think render will trigger if you set here `this.validators = this.validators.concat(validator)` as it's gonna create new array

src/renderer/containers/Dashboard/DashboardContainer.tsx
morrigan: Nop, can't access validators

mpetrunic: props.auth.account.getValidators().length? :)

morrigan: lol, that actually works xD

src/renderer/actions/network.ts
```diff
@@ -50,9 +51,7 @@ export const saveSelectedNetworkAction = (network: string): ISaveSelectedNetwork
 export const saveBeaconNodeAction = (url: string, network?: string) => {
     return async (dispatch: Dispatch<Action<unknown>>, getState: () => IRootState): Promise<void> => {
         const localDockerName = network ? BeaconChain.getContainerName(network) : undefined;
-        const signingKey = PrivateKey.fromBytes(
-            Buffer.from(getState().register.signingKey.replace("0x",""), "hex")
-        );
+        const signingKey = PrivateKey.fromBytes(fromHex(getState().register.signingKey.replace("0x","")));
```
mpetrunic: isn't from hex already removing 0x?

morrigan: No, that's just to create Buffer from string. We don't have much usage of this currently

mpetrunic: Can you add 0x removal there so we don't copy code everywhere

PR #183, opened 2020-03-19
Save validator network and managing Docker containers

morrigan:
Things done:

 - Saving validator network to database
 - Using spawn with detaching so Docker process is not killed together with CG
 - Starting stopped local beacon nodes `startAllLocalBeaconNodes` on login

### DIFF
--- a/src/renderer/actions/register.ts
+++ b/src/renderer/actions/register.ts
@@ -1,6 +1,6 @@
 import {RegisterActionTypes} from "../constants/action-types";
 import {Action, Dispatch} from "redux";
-import {ValidatorNetwork} from '../models/network';
+import {ValidatorNetwork} from "../models/network";
 import {IRootState} from "../reducers";
 import {V4Keystore} from "../services/keystore";
 import {Keypair} from "@chainsafe/bls/lib/keypair";
--- a/src/renderer/actions/register.ts
+++ b/src/renderer/actions/register.ts
@@ -164,9 +164,9 @@ export const addNewValidatorAction = (password: string) => {
         const state = getState();
 
         // Add new validator to database
-        await saveKeystore(signingKey, password),
+        await saveKeystore(signingKey, password);
         // Save validator's network
-        await saveNetwork(signingKey, state.register.network)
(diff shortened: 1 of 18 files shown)


src/renderer/actions/register.ts
```diff
@@ -116,12 +116,19 @@ const saveKeystore = async(state: IRootState, password: string): Promise<string>
     return accountDirectory;
 };
 
+const saveNetwork = async(signingKey: PrivateKey, networkName: string): Promise<void> => {
+    const network = new ValidatorNetwork(networkName);
+    const validatorAddress = signingKey.toPublicKey().toHexString();
```
mpetrunic: rename to validatorPubKey

src/renderer/actions/register.ts
```diff
@@ -149,9 +160,17 @@ export const completeRegistrationSubmission = (): Action<RegisterActionTypes> =>
...
     return async (dispatch: Dispatch<Action<unknown>>, getState: () => IRootState): Promise<void> => {
-        await saveKeystore(getState(), password);
+        const signingKey = PrivateKey.fromBytes(fromHex(getState().register.signingKey));
+        const state = getState();
+
+        await Promise.all([
+            // Add new validator to database
+            saveKeystore(signingKey, password),
+            // Save validator's network
+            saveNetwork(signingKey, state.register.network)
```
mpetrunic: I think we shouldn't do it at same time, saving keystore might fail afaik

src/renderer/containers/Login/LoginContainer.tsx
```diff
@@ -32,6 +33,10 @@ IOwnProps & IInjectedProps & Pick<IRootState, "auth">, IState> {
     public state: IState = {
         input: ""
     };
+
+    public async componentDidMount(): Promise<void> {
+        await BeaconChain.startAllLocalBeaconNodes();
```
mpetrunic: I would maybe move this to app container

morrigan: lol, makes more sense, not sure what I was thinking

PR #187, opened 2020-03-23
Beacon node syncing status

### LINKED ISSUE #180: Beacon node syncing status
Add status check (icon?!) to BN component that will display whether node is syncing or not.
See: https://api.prylabs.net/#/Node/GetSyncStatus

If node is syncing we should display current slot height instead of response time.
See: https://api.prylabs.net/#/BeaconChain/GetChainHead //headslot

morrigan:
Resolves #180


src/main/gui/window.ts
```diff
@@ -19,7 +19,7 @@ export async function createWindow(): Promise<void> {
     );
     // eslint-disable-next-line require-atomic-updates
     win = new BrowserWindow({
-        webPreferences: {nodeIntegration: true},
+        webPreferences: {nodeIntegration: true, webSecurity: false},
```
mpetrunic: ?

src/renderer/components/Cards/NodeCard.tsx
```diff
@@ -1,20 +1,35 @@
-import * as React from "react";
+import React from "react";
+import ReactTooltip from "react-tooltip";
 
 export interface INodeCardProps {
     title: string;
     value: number | string;
     url: string;
     onClick: () => void;
+    isSyncing?: boolean;
```
mpetrunic: Can value be -1 if not syncing to avoid this additional flag in interface?

src/renderer/components/Validator/ValidatorSimple.tsx
```diff
@@ -19,24 +22,45 @@ export interface IValidatorSimpleProps {
...
+    const dispatch = useDispatch();
+
+    useEffect(() => {
+        // Initially load beacon nodes
+        dispatch(loadValidatorBeaconNodes(props.publicKey));
+
+        // Periodically refresh status with latest height
+        const timeoutID = setInterval(() => {
+            dispatch(loadValidatorBeaconNodes(props.publicKey));
+        }, 5000);
```
mpetrunic: Can you add method in eth2ApiClient onNewHead. Some bn will have streaming api for that, as backup you can add this interval invoking :)

src/renderer/models/account.ts
```diff
@@ -60,8 +61,32 @@ export class CGAccount implements IAccount {
     /**
      * Returns array of beacon nodes that validator uses.
      */
-    public getValidatorBeaconNodes(validatorAddress: string): BeaconNode[] {
-        return this.validatorsBeaconNodes[validatorAddress];
+    public async getValidatorBeaconNodes(validatorAddress: string): Promise<BeaconNode[]> {
```
mpetrunic: I think this should return IEth2ApiClient and let container generate required data from that

src/renderer/services/eth2/client/prysm/prysm.ts
```diff
@@ -19,14 +23,35 @@ export class PrysmEth2ApiClient extends AbstractApiClient implements IValidatorB
 
     public constructor(options: IBeaconClientOptions) {
         super();
+        options.urlPrefix = `${options.urlPrefix}${apiPrefix}`;
```
mpetrunic: let's rename urlPrefix to baseUrl

src/renderer/services/eth2/client/prysm/prysm.ts
```diff
@@ -19,14 +23,35 @@ export class PrysmEth2ApiClient extends AbstractApiClient implements IValidatorB
...
         this.url = options.urlPrefix;
         this.config = options.config;
         this.beacon = new PrysmBeaconApiClient(options);
         this.validator = new PrysmValidatorApiClient(options);
     }
-    
+
+    public static getPrysmBeaconClient(url: string, network: string): PrysmEth2ApiClient | null {
```
mpetrunic: Can you create more universal method like
```javascript
getEth2ApiClient(url, network): IEth2ApiClient|null {
   switch(network) {
      case PRYSM: new PrysmEth2ApiClient({
            urlPrefix: url,
            config: networkConfig.eth2Config
        })
   }
}
```

src/renderer/services/eth2/client/prysm/prysm.ts
```diff
@@ -19,14 +23,35 @@ export class PrysmEth2ApiClient extends AbstractApiClient implements IValidatorB
...
+            urlPrefix: url,
+            config: networkConfig.eth2Config
+        });
+    }
+
     public async getVersion(): Promise<string> {
         return (await this.beacon.getClientVersion()).toString("ascii");
     }
 
+    public async isSyncing(): Promise<boolean> {
```
mpetrunic: Do you really need that util method

src/renderer/services/eth2/client/prysm/prysm.ts
```diff
@@ -19,14 +23,35 @@ export class PrysmEth2ApiClient extends AbstractApiClient implements IValidatorB
...
+        const result = await this.beacon.getSyncingStatus();
+        return (typeof result === "boolean") ? result : false;
+    }
+
+    public async getChainHeight(): Promise<string> {
```
mpetrunic: or this? haha

src/renderer/services/eth2/client/prysm/types.ts
```diff
@@ -46,4 +46,19 @@ export enum PrysmValidatorStatus {
     EXITING = "EXITING",
     SLASHING = "SLASHING",
     EXITED = "EXITED"
-}
\ No newline at end of file
+}
+
+export type ChainHead = {
```
mpetrunic: PrysmChainHeadResponse*

make sure IEth2Api returns our objects with appropriate types and conversions

src/main/gui/window.ts
morrigan: Enabling CORS requests. Prysm API doesn't support any config for that

src/renderer/components/Cards/NodeCard.tsx
morrigan: Yep!

src/renderer/services/eth2/client/prysm/prysm.ts
morrigan: This one seems to have 2 types of responses so yes :P

src/renderer/services/eth2/client/prysm/types.ts
morrigan: you mean here: https://github.com/NodeFactoryIo/ChainGuardian/blob/morrigan/bn-syncing-status/src/renderer/services/eth2/client/prysm/beacon.ts#L83 ?

src/main/gui/window.ts
mpetrunic: Can you open issue for that on prysmatic github and issue here(to remove this once they've implemented fix)?

src/renderer/services/eth2/client/prysm/prysm.ts
mpetrunic: Yeah but you can handle responses in container like
```javascript
const status = await api.beacon.getSyncingStatus()
if(!status) {
"synced"
} else {
"syncing"
}
```

src/renderer/services/eth2/client/prysm/types.ts
mpetrunic: yup, ChainHead should be type we use in our app, and PrysmChainHeadResponse is type we get from their api(byte arrays are hex encoded etc)

src/renderer/components/Validator/ValidatorSimple.tsx
morrigan: What should that method do?

mpetrunic: it subscription api it will execute param everytime bn processes block

src/renderer/services/eth2/client/prysm/types.ts
morrigan: Ok, but in this case they are the same.

src/renderer/components/Cards/NodeCard.tsx
morrigan: Actually, I am not sure how API works. What if it stops syncing but stops at some height? Currently we display either it is syncing and what's the current height. Otherwise, we can't tell if finished syncing or not.

mpetrunic: Well, you'll subscribe to new head to update slot number and you can check syncing status every time (or every other)

src/renderer/models/account.ts
morrigan: Can't do like that because of other beacon node info I need but will pull out something.

mpetrunic: Add required stuff to IEth2ApiClient?

mpetrunic: or include IEth2ApiClient in BeaconNode type

src/renderer/services/eth2/client/prysm/prysm.ts
```diff
@@ -23,35 +22,27 @@ export class PrysmEth2ApiClient extends AbstractApiClient implements IValidatorB
...
-    }
-
-    public async getChainHeight(): Promise<string> {
-        return (await this.beacon.getChainHead()).headSlot;
+    public onNewChainHead(callbacks: Array<any>, params: Array<any>): void {
+        // const eventSource = new EventSource(`${this.url}${PrysmBeaconRoutes.CHAINHEAD_STREAM}`);
+        // eventSource.onmessage = (e: MessageEvent): void => {
+        //     console.log("Got message!", e);
+        //     callbacks.map(callback => callback(e.data as PrysmChainHeadStreamResponse));
+        // };
```
mpetrunic: Did they confirm this doesn't work?

src/renderer/services/eth2/client/prysm/prysm.ts
```diff
@@ -23,35 +22,27 @@ export class PrysmEth2ApiClient extends AbstractApiClient implements IValidatorB
...
+
+        setInterval(async() => {
+            const response = await this.beacon.getChainHead();
+            callbacks.map(callback => callback.apply(this, [...params, response]));
+        }, 12000);
```
mpetrunic: config.params.SECONDS_PER_SLOT

src/renderer/services/eth2/client/prysm/prysm.ts
```diff
@@ -23,35 +22,27 @@ export class PrysmEth2ApiClient extends AbstractApiClient implements IValidatorB
...
     }
 
-    public async isSyncing(): Promise<boolean> {
-        const result = await this.beacon.getSyncingStatus();
-        return (typeof result === "boolean") ? result : false;
-    }
-
-    public async getChainHeight(): Promise<string> {
-        return (await this.beacon.getChainHead()).headSlot;
+    public onNewChainHead(callbacks: Array<any>, params: Array<any>): void {
```
mpetrunic: this should just accept single callback which you are suppose to persist in memory and call everytime there is new chainhead

src/renderer/services/eth2/client/prysm/types.ts
```diff
@@ -62,3 +62,11 @@ export type ChainHead = {
     previousJustifiedEpoch: string,
     previousJustifiedBlockRoot: string
 };
+
+export type PrysmChainHeadStreamResponse = {
+    result: PrysmChainHeadResponse,
+    error: {
+        httpStatus: string;
+        message: string;
+    }
```
mpetrunic: Are you sure that's response?

src/renderer/services/eth2/client/prysm/prysm.ts
morrigan: No, I didn't get any answer. However, I couldn't get it to work. It's possible that that is not working as I got some error regarding wrong MIME types and closed connections. I've only tried testing it through React using EventSource.

However, since you mentioned this is not available in Lighthouse, I figured no point in losing more time for the same thing.

src/renderer/services/eth2/client/prysm/types.ts
morrigan: Says so: https://api.prylabs.net/#/BeaconChain/StreamChainHead
Unfortunately I couldn't see for myself :D

src/renderer/services/eth2/client/prysm/prysm.ts
mpetrunic: Sure, because lighthouse will add those apis afaik

src/renderer/services/eth2/client/prysm/prysm.ts
morrigan: But that's what it does?

mpetrunic: it accepts array of <any> callbacks and params array.
You interface should be:
```javascript
public onNewChainHead(callback: (head: ChainHead) => void): void;
```

src/renderer/services/eth2/client/prysm/types.ts
mpetrunic: Oh, rename this to PrysmChainHeadStreamMessage since Chainhead is this:
https://api.prylabs.net/#/BeaconChain/GetChainHead

src/renderer/services/eth2/client/prysm/prysm.ts
morrigan: Why? And not to support params? :(

mpetrunic: callback should bind it's own context to encourage best practices and testing

src/config/config.ts
```diff
@@ -11,6 +11,9 @@ export interface IConfig {
     },
     db: {
         name: string
+    },
+    params: {
+        SECONDS_PER_SLOT: number
```
mpetrunic: wait no, you have this in IBeaconConfig.params.SECONDS_PER_SLOT

PR #191, opened 2020-03-29
Lighthouse api adapter

mpetrunic:


src/renderer/services/eth2/client/lighthouse/beacon.ts
```diff
@@ -0,0 +1,46 @@
+import {IBeaconApi} from "@chainsafe/lodestar-validator/lib/api/interface/beacon";
+import {bytes32, Fork, number64, SyncingStatus, uint64} from "@chainsafe/eth2.0-types";
+import {HttpClient} from "../../../api";
+import {IBeaconConfig} from "@chainsafe/eth2.0-config";
+import {IBeaconClientOptions} from "../interface";
+import {fromJson} from "@chainsafe/eth2.0-utils";
+
+export enum LighthouseBeaconRoutes {
+    VERSION = "/node/version",
```
morrigan: Maybe the ones that are same in both APIs should be in the same place? Will be easier to keep up and update if it gets standardized.

src/renderer/services/eth2/client/lighthouse/beacon.ts
```diff
@@ -0,0 +1,46 @@
...
+            chainId: BigInt(0)
+        };
+    }
+
+    public async getGenesisTime(): Promise<number64> {
+        return await this.client.get<number>(LighthouseBeaconRoutes.GENESIS_TIME);
+    }
+
+    public async getSyncingStatus(): Promise<boolean | SyncingStatus> {
+        return false;
```
morrigan: Why not implemented?

src/renderer/services/eth2/client/lighthouse/validator.ts
```diff
@@ -0,0 +1,123 @@
...
+            if(lhDuty) {
+                return {
+                    validatorPubkey: validatorPubKey,
+                    attestationSlot: lhDuty.attestation_slot,
+                    committeeIndex: lhDuty.attestation_committee_index
+                } as ValidatorDuty;
+            } else {
+                return null;
+            }
+        }).filter((value) => !!value) as ValidatorDuty[];
```
morrigan: Are you filtering out `null` values? It's hacky and confusing, rather not return null (not use map in that case)

src/renderer/services/eth2/client/lighthouse/beacon.ts
mpetrunic: routes or implementation?

src/renderer/services/eth2/client/lighthouse/beacon.ts
mpetrunic: In current version, they didn't implement it 😂

src/renderer/services/eth2/client/lighthouse/validator.ts
mpetrunic: What would you use?
This is somewhat common practicebI guess

src/renderer/services/eth2/client/lighthouse/beacon.ts
morrigan: wtf, how are we going to work without that? xD

src/renderer/services/eth2/client/lighthouse/beacon.ts
morrigan: routes

src/renderer/services/eth2/client/lighthouse/validator.ts
morrigan: class for loop? Idk, I am not sure why would iterating something twice be a common practice xD Plus, looks weird that you return something and then filter it.

mpetrunic: It will probably get inlined by node engine and I don't think it's gonna matter as it will mostly be 1 or 2 elem array :smile:

src/renderer/services/eth2/client/lighthouse/beacon.ts
mpetrunic: Well, it will never display syncing. :stuck_out_tongue: for now
I think they will implement it in targeted multi client version. You can always ask in their discord :)

src/renderer/services/eth2/client/lighthouse/beacon.ts
mpetrunic: I would keep them separated until api get's standardized because it will be hell to manage changes in api

src/renderer/services/eth2/client/lighthouse/validator.ts
morrigan: I was pointing to more readable and junior-friendly code rather than cooler code :P

PR #192, opened 2020-04-06
Close dev dockers

### LINKED ISSUE #184: Turn off docker containers on exit if in dev mode

morrigan:
Closes #184 

I don't like the code much so creating a draft so I can take a second look tomorrow :)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -62,7 +62,6 @@ module.exports = {
         "prefer-const": "error",
         "quotes": ["error", "double"],
         "semi": "off",
-        "react/prop-types": "off",
-        "no-async-promise-executor": "off"
+        "react/prop-types": "off"
     }
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -62,6 +62,7 @@ module.exports = {
         "prefer-const": "error",
         "quotes": ["error", "double"],
         "semi": "off",
-        "react/prop-types": "off"
+        "react/prop-types": "off",
+        "no-async-promise-executor": "off"
     }
 };
--- a/src/renderer/services/docker/chain.ts
+++ b/src/renderer/services/docker/chain.ts
@@ -40,25 +40,20 @@ export class BeaconChain extends Container {
     public static async startAllLocalBeaconNodes(): Promise<void> {
         const savedNodes = await database.beaconNodes.getAll();
         logger.info("Going to start all stopped local beacon nodes...");
-        const promises: any = [];
         for (let i = 0; i < savedNodes.length; i++) {
-            savedNodes[i].nodes.map((node) => {
+            savedNodes[i].nodes.map(async(node) => {
                 if (node.localDockerId) {
-                    const promise = new Promise(async(resolve, reject) => {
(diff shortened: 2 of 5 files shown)


.eslintrc.js
```diff
@@ -62,6 +62,7 @@ module.exports = {
         "prefer-const": "error",
         "quotes": ["error", "double"],
         "semi": "off",
-        "react/prop-types": "off"
+        "react/prop-types": "off",
+        "no-async-promise-executor": "off"
```
mpetrunic, line 66: I kinda like this, so you don't forget to await stuff :(

src/renderer/services/docker/chain.ts
```diff
@@ -44,7 +44,16 @@ export class BeaconChain extends Container {
...
-                    promises.push(Container.startStoppedContainer(node.localDockerId));
+                    const promise = new Promise(async(resolve, reject) => {
+                        const image = await Container.getImage(node.localDockerId);
+                        if (image) {
+                            await BeaconChain.loadStoppedChain(node.localDockerId, image);
+                        } else {
+                            return reject("Image not found.");
+                        }
+                        resolve();
+                    });
```
mpetrunic, line 55: Why don't you make function out this rather than some confusing promise ?

src/renderer/services/docker/chain.ts
```diff
@@ -56,6 +65,23 @@ export class BeaconChain extends Container {
...
+        }
+    }
+
+    private static async loadStoppedChain(name: string, image: string): Promise<void> {
+        const bc = new BeaconChain({
+            name,
+            image,
+        });
+        await bc.startStoppedContainer();
+        const network = BeaconChain.getNetworkFromContainerName(name);
```
mpetrunic, line 81: Maybe it would be better so save those connections in database?

src/renderer/services/docker/container.ts
```diff
@@ -45,19 +45,29 @@ export abstract class Container {
...
-    }
-
     public static async isContainerRunning(name: string): Promise<boolean> {
         const cmdResult = await runCmdAsync(Command.ps(name, "running"));
         // first line of output is header line, second line is definition of found docker instance
         const runningInstance = cmdResult.stdout.split("\n")[1];
         return runningInstance !== "";
     }
 
+    public static async getImage(name: string): Promise<string|undefined> {
```
mpetrunic, line 55: getImageName(dockerId: string)

.eslintrc.js
morrigan: No, this was something else completely but don't need it anymore

src/renderer/services/docker/chain.ts
morrigan: this wasn't ready for rreview :P

src/renderer/services/docker/chain.ts
morrigan: We have validator - network connection. But this is just docker image / beacon chain. I am not sure if we need this connection in any other way so didn't want to overcomplicate. We have a clear naming right now that we use it the network in container name.

PR #195, opened 2020-04-15
Disable multiple accounts

### LINKED ISSUE #176: Disable multiple account registrations

morrigan:
Resolves #176

### DIFF
--- a/src/renderer/services/utils/account.ts
+++ b/src/renderer/services/utils/account.ts
@@ -1,0 +1,24 @@
+import path from "path";
+import electron from "electron";
+import logger from "electron-log";
+
+import {getConfig} from "../../../config/config";
+import {V4Keystore} from "../keystore";
+import {removeDirRecursive} from "./file";
+
+export const cleanUpAccount = async(): Promise<void> => {
+    const config = getConfig(electron.remote.app);
+    try {
+        await Promise.all([
+            removeDirRecursive(config.storage.dataDir),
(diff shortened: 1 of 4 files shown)


src/renderer/services/utils/account.ts
```diff
@@ -0,0 +1,24 @@
+import path from "path";
+import electron from "electron";
+import logger from "electron-log";
+
+import {getConfig} from "../../../config/config";
+import {V4Keystore} from "../keystore";
+import {removeDirRecursive} from "./file";
+
+export const cleanUpAccount = async(): Promise<void> => {
+    const config = getConfig(electron.remote.app);
```
morrigan, line 10: @mpetrunic not sure if you noticed this one `app` passing. I guess it's safe to do? Can it be part of `getConfig` maybe?

mpetrunic: Isn't app suppose to be optional param, if it's not then it's fine

morrigan: It is optional but it gets `electron.app` which is used for main. For renderer that's undefined as it has `electron.remote.app`

mpetrunic: Not sure tbh, I think electron app is used for some specific configs so you might not need to pass anything

morrigan: Idk, was undefined. Works like this

PR #199, opened 2020-04-26
Update lodestar libraries

mpetrunic:
This is just PR to update libraries and comment out prysmatic client.

I decided to do upgrade on a specific branch, so I will open another PRs on this branch
with ligthhouse dockers and lighthouse client. I will also open up issues for some things like deposits, domain that changed and needs to be updated here


src/renderer/services/db/api/repositories/validator/attestations.ts
```diff
@@ -48,6 +48,6 @@ export class ValidatorAttestationsRepository extends BulkRepository<Attestation>
 
     private getAttestationKey(pubKey: BLSPubkey, attestation: Attestation): Buffer {
         const epoch = intToBytes(attestation.data.target.epoch, 2);
-        return Buffer.concat([pubKey, epoch, attestation.signature]);
+        return Buffer.concat([pubKey.valueOf() as Uint8Array, epoch, attestation.signature.valueOf() as Uint8Array]);
```
morrigan, line 51: What's with this return type of `valueOf` ?

mpetrunic: Underlying implementation is binary tree so valueOf() allows us to proxy tree and return it as Uint8Array. If it were structural backing, underlying representation is Uint8Array so valueOf() on Uint8Array is still Uint8Array.
You can read about that here: https://github.com/chainsafe/ssz#backings

PR #204, opened 2020-04-28
Update signature for DepositData to use computeSigningRoot

### LINKED ISSUE #201: Use DepositMessage and DepositData
ref: https://github.com/ethereum/eth2.0-specs/blob/v0.11.1/specs/phase0/validator.md#submit-deposit

morrigan:
Closes #201

### DIFF
--- a/src/renderer/services/deposit/utils.ts
+++ b/src/renderer/services/deposit/utils.ts
@@ -3,6 +3,7 @@ import {BLSPubkey as BLSPubKey, DepositData, DepositMessage} from "@chainsafe/lo
 import {createHash} from "crypto";
 import {config} from "@chainsafe/lodestar-config/lib/presets/mainnet";
 import {ethers, utils} from "ethers";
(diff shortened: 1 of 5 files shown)


src/renderer/services/deposit/utils.ts
```diff
@@ -3,6 +3,7 @@ import {BLSPubkey as BLSPubKey, DepositData, DepositMessage} from "@chainsafe/lo
 import {createHash} from "crypto";
 import {config} from "@chainsafe/lodestar-config/lib/presets/mainnet";
```
mpetrunic, line 4: Oh, we shouldn't assume mainnet config here

morrigan: I guess it was auto-import. Should use the one from `networks` ?

mpetrunic: yeah

PR #206, opened 2020-05-03
Spec 0.11.x

mpetrunic:


mpetrunic:
I would wait with merging until I've add script to test our validator against lighthouse

PR #210, opened 2020-05-08
Check custom beacon node

### LINKED ISSUE #207: Check if custom beacon URL is supported
- Check if it is working (ping version)
 - Check if supported (Lighthouse)

morrigan:
Closes #207


src/renderer/services/eth2/client/index.ts
```diff
@@ -27,3 +27,13 @@ export function getEth2ApiClient(url: string, network: string): IGenericEth2Clie
             });
     }
 }
+
+export async function isSupportedBeaconChain(url: string, network: string): Promise<boolean> {
+    const client = getEth2ApiClient(url, network);
+    try {
+        await client.beacon.getSyncingStatus();
```
mpetrunic, line 34: eh? we should get version. anc check if lighthouse in version. We can improve later by instantiating every supported client and see if it works

morrigan: Not sure what you want to check with version. The endpoint doesn't return "lighthouse" related version, not sure what do you mean. This is a hacky solution as it uses /lighthouse endpoint,specific to lighthouse.

mpetrunic: Wait, does it not return lighthouse+commit hash?
I don't think this is a good idea as this endpoint will soon be removed and we will have nodes with partial spec implementation so we will probably need to check build numbers and stuff

PR #214, opened 2020-05-13
Validator ROI & balance

morrigan:
Status (uptime) will be resolved with #213 

Also did some refactoring as ValidatorSimple is not simple anymore..


morrigan:
@mpetrunic Actually, if we don't need anything except balance from validator, I'll remove this additional redux reducer. I just realized that it might be redundant

src/renderer/actions/validator.ts
```diff
@@ -1,24 +1,59 @@
...
+}
+
+export const loadValidatorsAction = () => {
+    return async (dispatch: Dispatch<Action<unknown>>, getState: () => IRootState): Promise<void> => {
+        const auth = getState().auth;
+        if (auth && auth.account) {
+            const validators = auth.account.getValidators();
+            const validatorArray = validators.map((v) => ({
+                name: auth.account!.name,
+                status: "TODO status",
```
mpetrunic, line 21: This already exists, you should probably look into integrating in some future PR

src/renderer/actions/validator.ts
```diff
@@ -1,24 +1,59 @@
...
+                publicKey: v.publicKey.toHexString(),
+                deposit: 30,
```
mpetrunic, line 23: This can also be fetched from node instead of stubbed

src/renderer/actions/validator.ts
```diff
@@ -1,24 +1,59 @@
...
+
+            dispatch({
+                type: ValidatorActionTypes.LOAD_VALIDATORS,
+                payload: validatorArray,
+            });
+        }
+    };
+};
+
+export const loadValidatorBalance = (validator: string) => {
```
mpetrunic, line 36: Might be better if we fetch balance of multiple validators, which allows us to use single api request to fetch them all later on

src/renderer/actions/validator.ts
morrigan: I am not sure that we even need this info? Maybe we will when we support multiple deposits with values lower than 32 eth but currently this is not used and can remove it

src/renderer/actions/validator.ts
morrigan: Created #215

src/renderer/actions/validator.ts
mpetrunic: You can deposit more than 32 ethers though, but yeah, we can probably remove for now

src/renderer/components/Cards/ValidatorStat.tsx
```diff
@@ -33,20 +34,18 @@ const renderUptime = (props: IValidatorStatProps): React.ReactElement => {
...
 };
 export const ValidatorStat: React.FunctionComponent<IValidatorStatProps> = (
     props: IValidatorStatProps) => {
-    
+
     switch(props.type){
         case "ROI": return renderROI(props);
         case "ETH": return renderBalance(props);
-        case "Uptime": return renderUptime(props);
+        case "Status": return renderUptime(props);
```
mpetrunic, line 48: ?

morrigan: Uptime in design was by days so I guess this makes more sense?

PR #220, opened 2020-05-19
Validator logger

### LINKED ISSUE #179: Create logger for lodestar-validator
Logger should expose streams on which it will push logs (see https://www.npmjs.com/package/it-pushable).
End goal is to display those logs in existing react component for log streaming

morrigan:
Closes #179

### DIFF
--- a/src/renderer/services/eth2/client/logger.ts
+++ b/src/renderer/services/eth2/client/logger.ts
@@ -3,7 +3,7 @@ import {createLogger, format, Logger, transports} from "winston";
 import {defaultLogLevel, LogLevel, ILogger, ILoggerOptions} from "@chainsafe/lodestar-utils";
 import chalk from "chalk";
 
-export class ApiLogger implements ILogger {
+export class ValidatorLogger implements ILogger {
     private winston: Logger;
     private _level: LogLevel;
     private _silent: boolean;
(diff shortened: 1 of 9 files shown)


src/renderer/services/eth2/client/logger.ts
```diff
@@ -1,41 +1,123 @@
-import {ILogger, LogLevel} from "@chainsafe/lodestar-utils";
-import * as electronLogger from "electron-log";
+import {PassThrough} from "stream";
+import {createLogger, format, Logger, transports} from "winston";
+import {defaultLogLevel, LogLevel, ILogger, ILoggerOptions} from "@chainsafe/lodestar-utils";
+import chalk from "chalk";
 
 export class ApiLogger implements ILogger {
```
mpetrunic, line 6: ValidatorLogger?*

src/renderer/services/eth2/client/logger.ts
```diff
@@ -1,41 +1,123 @@
...
+            _level: options.level,
+            _silent: this._silent
+        });
+    }
+
+    public get stream(): PassThrough {
+        return this._stream;
+    }
+
+    private createLogEntry(level: LogLevel, message: string | object, context: object = {}): void {
```
mpetrunic, line 111: I see you've copied this from lodestar, might wanna check new implementation which fixes log levels. I'+ll take a look in winston logger so we expose method to pass custom tranfports

PR #236, opened 2020-05-23
GitHub actions

### LINKED ISSUE #224: Add ci for releasing of application binaries
Will probably require fixing windows build.
On tag, it will build and optionally sign binaries and upload them to release

mpetrunic:
- upgraded electron to 8.3.0
- removed screen tests from ci (it won't work reliably unless they start to maintain spectron)
- migrated from travis to github actions (tess will be run on only one os) - it will reduce build time
- added cd action which will on pushed tag build artifacts on windows, max and linux and upload those artifacts to github release

resolves #224


src/main/gui/window.ts
```diff
@@ -39,6 +44,7 @@ export async function createWindow(): Promise<void> {
         process.env.ELECTRON_DISABLE_SECURITY_WARNINGS = "1";
         win.loadURL("http://localhost:2003");
     } else {
+        console.log("setting route");
```
morrigan, line 47: logger use

mpetrunic, line 47:
```suggestion
```

mpetrunic:
> Also, what about code signing?

It will work automagically once we add base64 encoded certificate into secrets I think.
https://github.com/samuelmeuli/action-electron-builder#code-signing

Issue #234, opened 2020-05-22
Beacon node logs missing


mpetrunic:
![image.png](https://images.zenhubusercontent.com/5cd001edd503031def6241c6/5cb7aaf5-170b-49f5-b486-8e3048a894da)

morrigan:
This works only when running your own BN Docker.

mpetrunic:
I ran my own BN docker trough Chainguardian

morrigan:
> I ran my own BN docker trough Chainguardian

That should be fine... :/ Maybe it took a bit longer time to load?

mpetrunic:
Not sure, there was plenty of logs when I've queried Bn but Bn ran into errors. Maybe we should buffer logs before we get to that screen so we have something to display (like last 1k lines) and then just add to front and remove from end

PR #244, opened 2020-05-25
Fix starting beacon chain

morrigan:
Closes issues regarding to docker container not starting. From reported should be only #234


src/renderer/services/docker/container.ts
```diff
@@ -202,4 +214,17 @@ export abstract class Container {
         }
         return false;
     }
+
+    public async remove(): Promise<boolean> {
+        if (this.docker && this.docker.name) {
+            try {
+                runCmd(Command.removeContainer(this.docker.name));
+                logger.info(`Docker container ${this.docker.name} removed.`);
```
mpetrunic, line 222: delete volume?

src/renderer/services/utils/account.ts
```diff
@@ -26,6 +31,22 @@ export const deleteKeystore = (directory: string, publicKey: string): void => {
...
+            await DockerRegistry.removeContainerPermanently(network.dockerConfig.name);
+        }
+    }));
+
+    // Remove volumes
+    await Promise.all(networks.map(async(network) => {
+        if (network.networkName !== "localhost") {
+            await runCmdAsync(Command.removeVolume(network.dockerConfig.volumeName));
+        }
+    }));
```
mpetrunic, line 47: Probably makes more sense to do it inside container.remove?

src/renderer/services/docker/container.ts
morrigan: Maybe it shouldn't be coupled? In case you want to leave data, idk

mpetrunic: Well you cannot connect that data back anyways haha

src/renderer/services/utils/account.ts
```diff
@@ -26,6 +31,22 @@ export const deleteKeystore = (directory: string, publicKey: string): void => {
     selectedV4Keystore.destroy();
 };
 
+export const deleteBeaconNodeContainers = async(): Promise<void> => {
+    // Remove containers
+    await Promise.all(networks.map(async(network) => {
+        if (network.networkName !== "localhost") {
+            await DockerRegistry.removeContainerPermanently(network.dockerConfig.name);
```
mpetrunic, line 38: You should probably catch errors so we don't leave hanging containers in case of error

src/renderer/services/docker/container.ts
morrigan: Ok, wrong assumption then xD Will do

PR #261, opened 2020-06-16
Add beacon node

### LINKED ISSUE #257: Add beacon node screen
Should allow created validators to add new beacon node (display same configure screens as in onboarding)

morrigan:
Closes #257

### DIFF
--- a/src/renderer/containers/AddBeaconNode/AddBeaconNode.tsx
+++ b/src/renderer/containers/AddBeaconNode/AddBeaconNode.tsx
@@ -1,4 +1,4 @@
-import React, {useState} from "react";
+import React, {useState, useCallback} from "react";
 import {useDispatch} from "react-redux";
 import {useHistory, useParams} from "react-router";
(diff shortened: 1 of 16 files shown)


src/renderer/containers/AddBeaconNode/AddBeaconNode.tsx
```diff
@@ -0,0 +1,79 @@
...
+        return (
+            <InputBeaconNode
+                onGoSubmit={onGoSubmit}
+                onRunNodeSubmit={onRunNodeSubmit}
+            />
+        );
+    };
+
+    const renderSecondStep = (): React.ReactElement => {
+        const onDockerRunSubmit = (ports: DockerPort[], libp2pPort: string, rpcPort: string): void => {
```
mpetrunic, line 45: isn't this suppose to be useCallback so it get's recreated if history changes?

morrigan: Yeah, that would be optimization

PR #266, opened 2020-06-29
Normalize validators state

morrigan:
Improves performance while loading data where dashboard used to be lagging and blocked from clicking


src/renderer/actions/validator.ts
```diff
@@ -68,6 +68,7 @@ export const addNewValidator = (publicKey: string) => {
             network: getState().auth.account!.getValidatorNetwork(publicKey),
             keystore,
             status: undefined,
+            isRunning: undefined,
```
mpetrunic, line 71: why not false

PR #269, opened 2020-07-06
Add sentry and sourcemaps

### LINKED ISSUE #109: Enable sourcemap in webpack
- it should allow us to map bundled code with it's source for more accurate bug finding

### LINKED ISSUE #268: Setup Sentry for production build

morrigan:
Closes #109 
Closes #268

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,7 +2,7 @@
 name: ci
 
 # on events
-on: [push, pull_request]
+on: [pull_request]
 
 
 # jobs to run
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,7 +2,7 @@
 name: ci
 
 # on events
-on: [pull_request]
+on: [push, pull_request]
 
 
 # jobs to run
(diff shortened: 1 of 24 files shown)


.github/workflows/ci.yaml
```diff
@@ -2,7 +2,7 @@
 name: ci
 
 # on events
-on: [push, pull_request]
+on: [pull_request]
```
mpetrunic, line 5: this is open source we can afford to run it on psuh and pull_request :smile:

src/main/main.ts
```diff
@@ -3,6 +3,9 @@ import {initBLS} from "@chainsafe/bls";
 
 import {createWindow} from "./gui/window";
 import {DatabaseIpcHandler} from "./db/ipc";
+import {initSentry} from "./sentry";
+
+initSentry();
```
mpetrunic, line 8: I think we should leave user option on onboarding whether he would like to report those stuff or not. Since it could leak some privacy sensitive data

src/main/sentry.ts
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable @typescript-eslint/no-require-imports */
+export function initSentry(): void {
```
mpetrunic, line 2: you should probably put it outside of main as it's used in both?

morrigan: I think it should be in both

src/main/main.ts
morrigan: Yep, could be GDPR-ish too. Will finally get use of that Consent window

morrigan:
> Side question: Is this mapping to typescript files?

Yes

PR #272, opened 2020-07-20
Update lodestar libs v12

mpetrunic:
Note:
- switched to dummy eth1 setup because there is no command to make deposit for interop validators (we can either implement script for this ourselves or open issue on ligthhouse)
- connectivity with Lighthouse BN works partially (waiting for their response on what's wrong)


src/renderer/services/api/http/httpClient.ts
```diff
@@ -44,6 +44,7 @@ export class HttpClient {
 }
 
 const handleError = (error: AxiosError): Error => {
+    console.log(error);
```
morrigan, line 47: :D

PR #311, opened 2020-11-17
Implement custom validator name

### LINKED ISSUE #298: [Add validator] Post mnemonic load
After user enters private key/mnemonic, it should be asked to enter name (optional, autogenerated), password and confirm password, which will create keystore and lead user to dashboard with validator on top of list.

BeroBurny:
closes #298 

Screenshot from 2020-11-17 15-16-35


src/renderer/containers/Onboard/ConfigureValidator/ConfigureValidatorContainer.tsx
```diff
@@ -0,0 +1,58 @@
...
+        <>
+            <h1>Configure Validator settings</h1>
+            <p>You can skip customizing this data if you want to use the default values.</p>
+
+            <div className='configure-port'>
+                <div className='row'>
+                    <h3>Validator Name</h3>
+                </div>
+                <InputForm
+                    placeholder={`default: ${defaultName}`}
```
mpetrunic, line 45: I would remove "default:". I think it's fair to say everyone would get that stuff written there is gonna be applied as default value :smile:

src/renderer/containers/Onboard/ConfigureValidator/ConfigureValidatorContainer.tsx
```diff
@@ -0,0 +1,58 @@
...
+                capitalize(englishWordList[randBetween(0, englishWordList.length / 2)]) +
+                " " +
+                capitalize(englishWordList[randBetween(englishWordList.length / 2, englishWordList.length - 1)]),
+        );
+    }, []);
+
+    const dispatch = useDispatch();
+
+    const onSubmit = (): void => {
+        dispatch(setNameAction(name ? name : defaultName));
```
mpetrunic, line 31: Hmm I would rather set defaultName onChange. Because currently, if user would delete everything and clicked save it would silently save default name?

src/renderer/containers/Onboard/SigningKey/Import/FileUploadImport.tsx
```diff
@@ -53,7 +52,7 @@ export const FileUploadImport: FC<IOwnProps> = ({history}) => {
         if (isSwitching) {
             history.replace(Routes.ONBOARD_ROUTE_EVALUATE(OnBoardingRoutes.SIGNING_IMPORT_SLASHING_FILE));
         } else {
-            history.replace(Routes.ONBOARD_ROUTE_EVALUATE(OnBoardingRoutes.PASSWORD));
+            history.replace(Routes.ONBOARD_ROUTE_EVALUATE(OnBoardingRoutes.CONFIGURE));
```
mpetrunic, line 55: Didn't notice earlier. This will route user back to import screen if he clicks back?

src/renderer/ducks/register/slice.ts
```diff
@@ -67,6 +68,9 @@ export const registerSlice = createSlice({
         setPassword: (state, action: PayloadAction<string>): void => {
             state.password = action.payload;
         },
+        setName: (state, acton: PayloadAction<string>): void => {
+            state.name = acton.payload;
```
mpetrunic, line 72: I think you should update keystore as well?

src/renderer/containers/Onboard/SigningKey/Import/FileUploadImport.tsx
BeroBurny: worked kinda weird but fix it to be more intuitive

src/renderer/ducks/register/slice.ts
BeroBurny: here is stored temporarily so after can be stored at Keystore

mpetrunic (approved):

PR #314, opened 2020-11-23
Validator count on seed

### LINKED ISSUE #277: Display validator count when importing seed
To make it clearer, add disabled input field with number of validators (with value 1) and display pub key.

BeroBurny:
closes #277 

Screenshot from 2020-11-23 13-58-01

Screenshot from 2020-11-23 13-58-56

Screenshot from 2020-11-23 14-00-41


mpetrunic:
Index is not aligned (both context and field :smile: ). You should probably display just first x and last x letter of public key. Common practice in crypto is to remmber first or last letter so you can identify pubkey.

BeroBurny:
Screenshot from 2020-11-23 16-24-07

mpetrunic:
build failing, otherwise lgtm

morrigan (approved):

PR #325, opened 2020-12-01
Http client metrics

### LINKED ISSUE #5: Add metrics to http client
- track and store response times of each method of HTTP client
- add db method to calculate response avg over a given time period

This will be used in graphs. If a response is not received it should be saved as offline in uptime. Should watch for and write down next online status.

BeroBurny:
closes #5


src/renderer/models/networkLogs.ts
```diff
@@ -0,0 +1,42 @@
+export type NetworkLog = {
```
mpetrunic, line 1: Can you replace Log with Metrics in this file?

src/renderer/services/api/http/httpClient.ts
```diff
@@ -1,13 +1,16 @@
...
     private client: AxiosInstance;
+    private readonly loggerKey?: string;
 
-    public constructor(baseURL: string, options: {axios?: AxiosRequestConfig} = {}) {
+    public constructor(baseURL: string, options: {axios?: AxiosRequestConfig; loggerKey?: string} = {}) {
         if (!options) {
             // eslint-disable-next-line no-param-reassign
             options = {axios: {}};
         }
+        this.loggerKey = options.loggerKey;
```
mpetrunic, line 13: I think loggerKey should be baseURL?

BeroBurny: it could be but is not exclusive to be baseURL
like in case if you want record metrics for every validator, but these validators use the same beacon node, whit baseURl you will be unable to separate results

src/renderer/models/networkMetrics.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public addRecord(record: NetworkMetric): void {
+        this.records.push(record);
+        this.prune();
+    }
+
+    // remove logs older that 1 day
+    private prune(): void {
+        const dayBefore = Date.now() - 25 * 60 * 60 * 1000;
```
mpetrunic, line 39: magic numbers, extract to constant

src/renderer/models/types/networkMetrics.ts
```diff
@@ -0,0 +1,20 @@
...
+        url: new StringType(),
+        code: new NumberUintType({byteLength: 32}),
+        latency: new NumberUintType({byteLength: 32}),
+        time: new NumberUintType({byteLength: 32}),
+    },
+});
+
+export const NetworkMetricsType = new ContainerType<NetworkMetrics>({
+    fields: {
+        records: new ListType({elementType: NetworkMetricType, limit: MILLION}),
```
mpetrunic, line 18: probably better to create unboundedArray, with serialize and json methods implemented however you wan't. Milion might not be enough.

BeroBurny: I wanted to bu limit is a required field, that why I put a million hoping it will be enoight

src/renderer/services/api/http/httpClient.ts
mpetrunic: hmm, but validator metrics are not collected here? We collect beacon node metrics here which is uniquely identified by url

src/renderer/models/types/networkMetrics.ts
mpetrunic: I've answered this issue in another PR

src/renderer/services/api/http/httpClient.ts
BeroBurny: HTTP client is not strictly linked to beacon, can be used in validator too or any other HTTP requests

a possible problem with the baseURL, you can have validator on `localhost:5052` so the baseURL will be `http://localhost:5052/eth/v1/` or even `http://localhost:5052/eth/v1/beacon` or there is a new version of APIs so old logs will be "forgotten"

src/renderer/models/types/networkMetrics.ts
```diff
@@ -0,0 +1,18 @@
...
+        url: new StringType(),
+        code: new NumberUintType({byteLength: 32}),
+        latency: new NumberUintType({byteLength: 32}),
+        time: new NumberUintType({byteLength: 32}),
+    },
+});
+
+export const NetworkMetricsType = new ContainerType<NetworkMetrics>({
+    fields: {
+        records: new CompositeArrayType({elementType: NetworkMetricType}),
```
mpetrunic, line 16: Don't think this works.
CompositeArrayType doesn't have deserialize/serialize correctly implemented.

BeroBurny: I can return back to List and open an issue to optimize this?

BeroBurny: the issue at #348

mpetrunic (approved):

PR #326, opened 2020-12-01
implement accordion

### LINKED ISSUE #319: Collapse advanced settings in "Configure Beacon node" screen

BeroBurny:
closes #319 

Screenshot from 2020-12-01 17-42-09

Screenshot from 2020-12-01 17-42-14

### DIFF
--- a/src/renderer/components/Accordion/Accordion.tsx
+++ b/src/renderer/components/Accordion/Accordion.tsx
@@ -1,0 +1,28 @@
+import React, {useEffect, useRef} from "react";
+
+interface IProps {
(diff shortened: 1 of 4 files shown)


morrigan (commented):
Chain data location should be ChainGuardian storage path or so? Maybe should be an advanced setting as well?
Also, would be better to put note next to ETH1 endpoint that we strongly suggest registering on Infura and using own product key for reliable performance.

BeroBurny:
@morrigan @mpetrunic :eyes:

morrigan (approved):
Fine by me! :)

PR #332, opened 2020-12-03
Refactor schema validator

### LINKED ISSUE #330: Decide between JOI and AJV

BeroBurny:
closes #330


mpetrunic (commented):
Oh man, I didn't see you did this. We already have that implemented in new lodestar I think.
https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar-validator/src/slashingProtection/interchange/parseInterchange.ts

mpetrunic (commented):
Hmm I think we should sue validation stuff from ldoestar-validator:
https://github.com/ChainSafe/lodestar/blob/4ca5a760c928c1257172bb4bf1033de2f60e6f3d/packages/lodestar-validator/src/slashingProtection/interchange/parseInterchange.ts#L8

as it provides better checks (config, and genesis root should be pulled from network config).

We should probably revisit this since we don't know network at this point?

morrigan:
We can merge this and @mpetrunic will create issue for importing and validating slashing db

morrigan (approved):

PR #352, opened 2020-12-16
Set beacon node docker default to home folder

### LINKED ISSUE #342: Set default path of beacon chain data to user specific aplication directory with beacon subdir

BeroBurny:
closes #342

### DIFF
--- a/src/renderer/components/ConfigureBeaconNode/ConfigureBeaconNode.tsx
+++ b/src/renderer/components/ConfigureBeaconNode/ConfigureBeaconNode.tsx
@@ -28,7 +28,7 @@ export const ConfigureBeaconNode: React.FunctionComponent<IConfigureBNProps> = (
     const [networkIndex, setNetworkIndex] = useState(0);
     const defaults = getDefaultsForClient(props.clientName);
 
-    const defaultChainDataDir = path.join(getConfig(remote.app).storage.home) + "/beacon";
+    const defaultChainDataDir = path.join(getConfig(remote.app).storage.home, "beacon");
     const [chainDataDir, setChainDataDir] = useState(defaultChainDataDir);
 
     const defaultEth1URL = "https://goerli.infura.io/v3/73d4045fe98a406faa2334ad7306b313";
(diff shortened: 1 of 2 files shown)


src/renderer/components/ConfigureBeaconNode/ConfigureBeaconNode.tsx
```diff
@@ -28,7 +28,7 @@ export const ConfigureBeaconNode: React.FunctionComponent<IConfigureBNProps> = (
     const [networkIndex, setNetworkIndex] = useState(0);
     const defaults = getDefaultsForClient(props.clientName);
 
-    const defaultChainDataDir = path.join(getConfig(remote.app).storage.dataDir);
+    const defaultChainDataDir = path.join(getConfig(remote.app).storage.home) + "/beacon";
```
mpetrunic, line 31: I don't think we should store chain data in ~/beacon. It's wrong on so many levels 😄 
I think according to xdg, you should use `path.join(getConfig(remote.app).storage.dataDir, "beacon");`

https://www.electronjs.org/docs/api/app#appgetpathname

mpetrunic (changes requested):

src/renderer/components/ConfigureBeaconNode/ConfigureBeaconNode.tsx
```diff
@@ -28,7 +28,7 @@ export const ConfigureBeaconNode: React.FunctionComponent<IConfigureBNProps> = (
     const [networkIndex, setNetworkIndex] = useState(0);
     const defaults = getDefaultsForClient(props.clientName);
 
-    const defaultChainDataDir = path.join(getConfig(remote.app).storage.dataDir);
+    const defaultChainDataDir = path.join(getConfig(remote.app).storage.home, "beacon");
```
mpetrunic, line 31:
```suggestion
    const defaultChainDataDir = path.join(getConfig(remote.app).storage.dataDir, "beacon");
```

mpetrunic (changes requested):

src/renderer/components/ConfigureBeaconNode/ConfigureBeaconNode.tsx
BeroBurny: so are we gonna abandon this task?

mpetrunic: I think we still need beacon subdir

mpetrunic (approved):

PR #365, opened 2020-12-22
Fix validator info

### LINKED ISSUE #361: Fix default ROI
Should not be `-100%` :D

### LINKED ISSUE #363: Balance/ROI/Status

BeroBurny:
closes #363 
closes #361 
additional refactor on chart data that I forger in the last pr


morrigan (approved):
LGTM

src/renderer/components/Cards/ValidatorStat.tsx
```diff
@@ -4,29 +4,37 @@ import {utils} from "ethers";
...
-                {props.value >= 0 ? "+" : ""}
-                {props.value}%
-            </h1>
-            <h5 className={props.value >= 0 ? "plus" : "minus"}>{props.type}</h5>
+            {props.value === undefined ? (
+                <h1>N/A</h1>
+            ) : (
+                <h1 className={props.value >= 0 ? "plus" : "minus"}>
+                    {props.value >= 0 ? "+" : ""}
+                    {props.value}%
```
mpetrunic, line 18: shoudln't this be formatted to be rounded to two decimals?

src/renderer/ducks/validator/sagas.ts
```diff
@@ -62,13 +50,16 @@ function* loadValidatorsSaga(): Generator<
...
                 return {
                     name: keyStore.getName() ?? `Validator - ${index}`,
                     status: undefined,
                     publicKey: keyStore.getPublicKey(),
-                    network: auth.getValidatorNetwork(keyStore.getPublicKey()),
+                    network,
+                    balance,
                     keystore: keyStore,
-                    isRunning: undefined,
+                    isRunning: false,
```
mpetrunic, line 62: Aren't you suppose to take this value depending on if lodestar-validator is running or not?

BeroBurny: Definitely loadstar validator is not running, so seeing no point to keep status undefined if we are sure he is not started and another case `undefined` is a falsy value at the end is same result

src/renderer/components/Cards/ValidatorStat.tsx
BeroBurny: dunno just extended to display `N/A` if there is no value

mpetrunic: let's format it to be on safe side

src/renderer/ducks/validator/sagas.ts
mpetrunic: do we change it somewhere to true?

BeroBurny: after starting a validator

src/renderer/components/Cards/ValidatorStat.tsx
BeroBurny: i check now and yeah its formatted to 2 decimals at `calculateROI` function
https://github.com/NodeFactoryIo/ChainGuardian/blob/78743d0b0d429ea60e5c2745d081274762cb9c67/src/renderer/services/utils/math.ts#L16

mpetrunic (approved):

PR #369, opened 2020-12-23
Import slashing db

### LINKED ISSUE #343: Use lodestar-validator slashing db interchange validation
> Hmm I think we should use validation stuff from ldoestar-validator:
https://github.com/ChainSafe/lodestar/blob/4ca5a760c928c1257172bb4bf1033de2f60e6f3d/packages/lodestar-validator/src/slashingProtection/interchange/parseInterchange.ts#L8

as it provides better checks (config, and genesis root should be pulled from network config).

We should remove slashing db import from create validator and add it on start of validator client at which point we are aware of validator network.

BeroBurny:
closes #343


src/renderer/ducks/validator/sagas.ts
```diff
@@ -113,11 +136,48 @@ function* startService(
...
+        const {genesisValidatorsRoot} = yield eth2API.beacon.getGenesis();
+        const interchange: InterchangeFormatVersion = {
+            format: "complete",
+            version: "4",
+        };
+        const validatorId = new Uint8Array(Buffer.from(publicKey.substr(2), "hex"));
+        slashingProtection.exportInterchange(genesisValidatorsRoot, [validatorId], interchange).then((data) => {
+            writeFileSync(path, JSON.stringify(data));
+        });
+        throw new Error("Sucker!");*/
```
mpetrunic, line 179: lol, remove comments

mpetrunic (commented):
lgtm, just 1 minor thing

mpetrunic (approved):

PR #370, opened 2020-12-23
Validator statuses

### LINKED ISSUE #340: Update validator status script
- with new changes to eth2 api validator endpoint now contains validator status
- we should map those statuses to our descriptions (update existing code)
- if endpoint doesn't return validator, we should use eth1 node to determine validator status

BeroBurny:
closes #340

### DIFF
--- a/src/renderer/services/utils/getValidatorState.ts
+++ b/src/renderer/services/utils/getValidatorState.ts
@@ -15,7 +15,7 @@ export const getValidatorState = async (publicKey: string, beaconNodeUrl: string
 
     // @ts-ignore
     if (stateValidator === null || stateValidator.status === "unknown") {
-        const depositInfo = await client.beaconConfig.getDepositContract();
+        const depositInfo = await client.networkConfig.getDepositContract();
 
         const genesis = await client.beacon.getGenesis();
         const genesisForkVersion = "0x" + Buffer.from(genesis.genesisForkVersion as Uint8Array).toString("hex");
--- a/src/renderer/services/utils/getValidatorState.ts
+++ b/src/renderer/services/utils/getValidatorState.ts
@@ -1,0 +1,94 @@
+import {ValidatorStatus} from "../../constants/validatorStatus";
+import {CgEth2ApiClient} from "../eth2/client/eth2ApiClient";
+import {config as mainnetConfig} from "@chainsafe/lodestar-config/lib/presets/mainnet";
+import {getNetworkConfigByGenesisVersion} from "../eth2/networks";
+import {ethers} from "ethers";
+
+export const getValidatorState = async (publicKey: string, beaconNodeUrl: string): Promise<ValidatorStatus> => {
+    if (!beaconNodeUrl) return ValidatorStatus.NO_BEACON_NODE;
+
+    // TODO: get config from based on network
+    const client = new CgEth2ApiClient(mainnetConfig, beaconNodeUrl);
+
+    const validatorId = new Uint8Array(Buffer.from(publicKey.substr(2), "hex"));
+    const stateValidator = await client.beacon.state.getStateValidator("head", validatorId);
+
+    // @ts-ignore
+    if (stateValidator === null || stateValidator.status === "unknown") {
+        const depositInfo = await client.beaconConfig.getDepositContract();
+
+        const genesis = await client.beacon.getGenesis();
+        const genesisForkVersion = "0x" + Buffer.from(genesis.genesisForkVersion as Uint8Array).toString("hex");
+        const config = getNetworkConfigByGenesisVersion(genesisForkVersion);
+
+        // TODO: experiment to see best
+        const currentBlock = await config.eth1Provider.getBlockNumber();
+        const logs = await config.eth1Provider.getLogs({
+            address: depositInfo.address,
+            fromBlock: currentBlock - 6144, // approx 24h before current block
+            topics: [ethers.utils.id("DepositEvent(bytes,bytes,bytes,bytes,bytes)")],
+        });
+        const containsThisValidator = logs.some(({data}) => data.includes(publicKey.substr(2)));
+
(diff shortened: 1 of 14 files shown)


src/renderer/services/utils/getValidatorState.ts
```diff
@@ -0,0 +1,96 @@
...
+        const config = getNetworkConfigByGenesisVersion(genesisForkVersion);
+
+        // TODO: experiment to see best
+        const currentBlock = await config.eth1Provider.getBlockNumber();
+        const logs = await config.eth1Provider.getLogs({
+            address: depositInfo.address,
+            fromBlock: currentBlock - 6144, // approx 24h before current block
+            topics: [ethers.utils.id("DepositEvent(bytes,bytes,bytes,bytes,bytes)")],
+        });
+        const containsThisValidator = logs.some(({data}) => data.includes(publicKey.substr(2)));
```
mpetrunic, line 31: extract this chunk to separate method

src/renderer/services/utils/getValidatorState.ts
```diff
@@ -0,0 +1,96 @@
...
+                return ValidatorStatus.SLASHED;
+            // @ts-ignore
+            case "withdrawable":
+                return ValidatorStatus.WITHDRAWABLE;
+            // @ts-ignore
+            case "withdrawn":
+                return ValidatorStatus.WITHDRAWNED;
+            default:
+                console.error(`Status: "${stateValidator.status}" not found`);
+                return ValidatorStatus.ERROR;
```
mpetrunic, line 93: extract switch-case to separate method

src/renderer/services/utils/getValidatorState.ts
```diff
@@ -0,0 +1,96 @@
...
+                return ValidatorStatus.ACTIVE;
+            case "active_awaiting_voluntary_exit":
+                return ValidatorStatus.GOOD_BOY_EXITING;
+            case "active_awaiting_slashed_exit":
+                return ValidatorStatus.SLASHED_EXITING;
+            case "exited_voluntarily":
+                return ValidatorStatus.VOLUNTARILY_EXITED;
+            case "exited_slashed":
+                return ValidatorStatus.SLASHED;
+            // @ts-ignore
```
mpetrunic, line 85: why ts-ignore? You should probably extend Validator type to be LighthouseValidator with fields and statuses it actually returns

mpetrunic (changes requested):

src/renderer/services/utils/getValidatorState.ts
```diff
@@ -0,0 +1,96 @@
+import {ValidatorStatus} from "../../constants/validatorStatus";
+import {CgEth2ApiClient} from "../eth2/client/eth2ApiClient";
+import {config as mainnetConfig} from "@chainsafe/lodestar-config/lib/presets/mainnet";
+import {getNetworkConfigByGenesisVersion} from "../eth2/networks";
+import {ethers} from "ethers";
+
+export const getValidatorState = async (publicKey: string, beaconNodeUrl: string): Promise<ValidatorStatus> => {
+    if (!beaconNodeUrl) return ValidatorStatus.NO_BEACON_NODE;
+
+    // TODO: get config from based on network
```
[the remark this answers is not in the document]
BeroBurny: we do not have but think we need change approach for getting a config

src/renderer/services/utils/getValidatorState.ts
```diff
@@ -0,0 +1,100 @@
...
+        const config = getNetworkConfigByGenesisVersion(genesisForkVersion);
+
+        // TODO: experiment to see best
+        const currentBlock = await config.eth1Provider.getBlockNumber();
+        const logs = await config.eth1Provider.getLogs({
+            address: depositInfo.address,
+            fromBlock: currentBlock - 6144, // approx 24h before current block
+            topics: [ethers.utils.id("DepositEvent(bytes,bytes,bytes,bytes,bytes)")],
+        });
+        const containsThisValidator = logs.some(({data}) => data.includes(publicKey.substr(2)));
```
mpetrunic, line 33: Don't use substr

src/renderer/services/utils/getValidatorState.ts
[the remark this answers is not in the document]
mpetrunic: Hmm can you make standalone method to fetch genesis root (without using ssz). We can use that to determine config and init eth2Client?

src/renderer/services/utils/getValidatorState.ts
mpetrunic: Also it's incorrect since validator can deposit less than 32Eth in which case status isn't PROCESSING_DEPOSIT but it has logs. Deposit could also be invalid.

To avoid all that, I would skip eth1 stuff for now. If it's missing in validator registry, just set PENDING_DEPOSIT and we can improve that later on

src/renderer/services/utils/getValidatorState.ts
```diff
@@ -0,0 +1,100 @@
+import {ValidatorStatus} from "../../constants/validatorStatus";
+import {CgEth2ApiClient} from "../eth2/client/eth2ApiClient";
+import {config as mainnetConfig} from "@chainsafe/lodestar-config/lib/presets/mainnet";
+import {getNetworkConfigByGenesisVersion} from "../eth2/networks";
+import {ethers} from "ethers";
+import {fromHexString} from "@chainsafe/ssz";
+import {toHex} from "@chainsafe/lodestar-utils";
+
+export const getValidatorState = async (publicKey: string, beaconNodeUrl?: string): Promise<ValidatorStatus> => {
```
mpetrunic, line 9:
```suggestion
export const getValidatorStatus = async (publicKey: string, beaconNodeUrl?: string): Promise<ValidatorStatus> => {
```
rename file as well

mpetrunic (changes requested):
I think you need to remove old method for getting validator status?!

mpetrunic (approved):

PR #391, opened 2021-01-05
Validator balance chart

### LINKED ISSUE #324: Validator balance graph

BeroBurny:
closes #324 

Screenshot from 2021-01-05 09-16-43


mpetrunic:
What are those sharp balance drops?

BeroBurny:
> What are those sharp balance drops?

its messed with historical and current data, dint flush DB before taking screenshot :sweat: 

> Can we hint data types on x and y axis?

there is hover with formatted ETH value but cannot take if on screenshot :disappointed:

mpetrunic:
> there is hover with formatted ETH value but cannot take if on screenshot
That's fine, but can we add label to axis so it's displays "Pages"/"Index" (In our example it could be "Balance Gwei" / "Epoch")  like here:
https://recharts.org/en-US/examples/ComposedChartWithAxisLabels

mpetrunic (approved):

PR #401, opened 2021-01-07
update validator balance on head

### LINKED ISSUE #393: Update validator balance on head epoch change rather than finalized epoch change

BeroBurny:
closes #393


src/renderer/ducks/beacon/sagas.ts
```diff
@@ -222,17 +206,20 @@ export function* watchOnHead(
...
-                yield put(updateSlot(payload.value.message.slot, url));
-                if (isSyncing || !isOnline) {
-                    const result = yield client.node.getSyncingStatus();
-                    isSyncing = result.syncDistance > 10;
-                    isOnline = true;
-                    yield put(updateStatus(isSyncing ? BeaconStatus.syncing : BeaconStatus.active, url));
-                }
+
+            yield put(updateSlot(payload.value.message.slot, url));
+            const headEpoch = Math.floor(payload.value.message.slot / 32);
```
mpetrunic, line 211: there is method for that in state transition package:
https://github.com/ChainSafe/lodestar/blob/master/packages/lodestar-beacon-state-transition/src/util/epoch.ts#L13

mpetrunic (changes requested):
you should use epoch instead of head when fetching validator balance update

src/renderer/ducks/validator/sagas.ts
```diff
@@ -234,18 +234,23 @@ function* validatorInfoUpdater(
...
             }
 
             const validator = yield select(getValidator, {publicKey});
             if (validator.beaconNodes.includes(payload.beacon)) {
-                const balance = yield getValidatorBalance(publicKey, network, payload.beacon);
+                const balance = yield getValidatorBalance(
+                    publicKey,
+                    network,
+                    payload.beacon,
+                    BigInt(payload.epoch * 32),
```
mpetrunic, line 252: ugh, heck:
- I totally forgot validator is fetched by slot (don't convert to epoch and back to slot here, just send slot here)
- slot isn't BigInt

mpetrunic (changes requested):

mpetrunic (approved):

PR #402, opened 2021-01-07
Fix env variables and sentry source upload

### LINKED ISSUE #399: Generate and upload sourcemaps to Sentry

mpetrunic:
- add env variable replacement during build
- add sentry plugin for sourcemap upload
- remove type and eslint checking in webpack

@morrigan Can you take a detail look and set required secrets.

resolves #399


morrigan:
I've already set the secrets because using them in my draft PR so check to align that.

mpetrunic:
> I've already set the secrets because using them in my draft PR so check to align that.

there is only sentry dns from what I saw

webpack.base.config.js
```diff
@@ -30,6 +30,9 @@ module.exports = {
         new Dotenv(),
         new webpack.IgnorePlugin({
             resourceRegExp: /\@chainsafe\/blst/,
-          })
+          }),
+          new webpack.DefinePlugin({
+            "process.env.DOCKER_LIGHTHOUSE_IMAGE": JSON.stringify(process.env.DOCKER_LIGHTHOUSE_IMAGE || "sigp/lighthouse:v1.0.5")
```
BeroBurny, line 35: better use `sigp/lighthouse:latest` verzion `1.0.5` could be deprecated in few months

BeroBurny:
probably this package will help in maintenance https://www.npmjs.com/package/dotenv-webpack

.github/workflows/release.yaml
```diff
@@ -110,6 +110,11 @@ jobs:
           mac_certs_password: ${{ secrets.MAC_PASSWORD }}
         env:
           EP_PRE_RELEASE: true
+          SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
+          SENTRY_ORG: "nodefactory"
+          SENTRY_PROJECT: "chainguardian"
+          SENTRY_SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
```
morrigan, line 116: this is not used, can be removed

webpack.base.prod.config.js
```diff
@@ -1,15 +1,35 @@
...
+        "process.env.SENTRY_AUTH_TOKEN": JSON.stringify(process.env.SENTRY_AUTH_TOKEN || ""),
+    }),
+];
+
+if(process.env.SENTRY_AUTH_TOKEN) {
+    // Sentry CLI used for uploading releases
+    plugins.push(new SentryWebpackPlugin({
+
+        // sentry-cli configuration
+        authToken: process.env.SENTRY_AUTH_TOKEN,
```
morrigan, line 19: I see now. You replace DSN usage with this?

mpetrunic: this is for sourcemap upload (private value), dsn is for error reporting (public value)

.github/workflows/release.yaml
mpetrunic: it's used by webpack during release process to upload sourcemap

morrigan (approved):
Looks good. Added auth secret

webpack.base.config.js
morrigan: ha, double edged sword :D
We can let users do "docker pull" if update is needed. To me looks better to be on stable version that's working for us

PR #456, opened 2021-01-27
Autostart validator

### LINKED ISSUE #450: Auto-start validator on active state
When the user creates a new validator Keystore, import it to ChainGuartian add beacon node then user need to be able to start validator automatically when the validator become active for work on eth2

BeroBurny:
closes #450 
blocker #443


mpetrunic (commented):
This also resolves validator export?

BeroBurny:
> This also resolves validator export?

nope it is branched from #443

mpetrunic (approved):

PR #475, opened 2021-02-01
Generate changelog

### LINKED ISSUE #444: Add generating changelog from PR titles
Example: https://github.com/ChainSafe/lodestar/blob/master/.github/workflows/release.yml#L66

BeroBurny:
closes #444

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -26,61 +26,6 @@ jobs:
           with_initial_version: false
         id: get-latest-tag
 
-    outputs:
-      tag: ${{ steps.tag.outputs.tagname }}
-      previous_tag: ${{ steps.get-latest-tag.outputs.tag }}
-      version: ${{ steps.tag.outputs.version }}
-
-  release:
-    name: Publish
-    runs-on: ${{ matrix.os }}
-    needs: tag
-    strategy:
-      matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -114,15 +80,44 @@ jobs:
...
+          author: "true"
+          usernamesAsGithubLogins: "true"
+          compareLink: "true"
+          filterByMilestone: "false"
+          unreleased: "false"
+          sinceTag: "${{ needs.tag.outputs.previous_tag }}"
+          maxIssues: "0"
+          stripGeneratorNotice: "true"
+
+      - name: Create Release
```
mpetrunic, line 102: This cannot be after `name: Build/release Electron app` since we wouldn't have release artifacts then. Also this job is executed on every os, which means we would create release 3 times. You should leave everything as it was but add generate changelog in "tag" job so the flow would be following:
Check and Tag:
1. check if new tag and tag
2. create changelog
3. create release
4. (rollback on failure)

Publish:
1. yarn install
2. build and upload release

So tldr:
Just add generate changelog and get previous version to tag job

mpetrunic (changes requested):

mpetrunic (approved):

PR #487, opened 2021-02-03
Docker performance monitor

### LINKED ISSUE #480: Hardware metrics
We should display hardware metrics for beacon nodes that are runned locally through Docker. We can use `docker stats` to obtain the information (CPU, Memory usage, Network) real time. Ideally, maybe we can use an API for it? Otherwise, if no API, we should parse the log output.

BeroBurny:
closes #480


mpetrunic (commented):
There is a lot of warnings that stats aren't covered with unit tests so it would be nice if you could test that

mpetrunic (approved):

PR #493, opened 2021-02-09
Fix atterstation effectivnes

### LINKED ISSUE #485: Reorganize validator and beacon node tabs
When beacon node is clicked, only beacon node information should be displayed:

 * beacon node details  - "Dashboard"
 * beacon node logs - "Logs"
 * performance - "Performance"

Remove beacon node from validator details.

BeroBurny:
closes #485

### DIFF
--- a/src/renderer/ducks/validator/actions.ts
+++ b/src/renderer/ducks/validator/actions.ts
@@ -83,15 +83,15 @@ export type SignedNewAttestation = (
     block: string,
     index: number,
     slot: number,
-    bitsIndex: number,
+    validatorIndexInCommittee: number,
 ) => {
-    payload: {block: string; index: number; slot: number; bitsIndex: number};
+    payload: {block: string; index: number; slot: number; validatorIndexInCommittee: number};
     meta: string;
 };
 export const signedNewAttestation = createAction<SignedNewAttestation>(
     "validator/signedNewAttestation",
-    (publicKey: string, block: string, index: number, slot: number, bitsIndex: number) => ({
-        payload: {block, index, slot, bitsIndex},
+    (publicKey: string, block: string, index: number, slot: number, validatorIndexInCommittee: number) => ({
+        payload: {block, index, slot, validatorIndexInCommittee},
         meta: publicKey,
     }),
 );
--- a/src/renderer/ducks/validator/actions.ts
+++ b/src/renderer/ducks/validator/actions.ts
@@ -83,14 +83,15 @@ export type SignedNewAttestation = (
     block: string,
     index: number,
     slot: number,
+    bits: string,
 ) => {
-    payload: {block: string; index: number; slot: number};
+    payload: {block: string; index: number; slot: number; bits: string};
     meta: string;
 };
 export const signedNewAttestation = createAction<SignedNewAttestation>(
     "validator/signedNewAttestation",
-    (publicKey: string, block: string, index: number, slot: number) => ({
-        payload: {block, index, slot},
+    (publicKey: string, block: string, index: number, slot: number, bits: string) => ({
+        payload: {block, index, slot, bits},
         meta: publicKey,
     }),
 );
--- a/src/renderer/ducks/validator/actions.ts
+++ b/src/renderer/ducks/validator/actions.ts
@@ -83,15 +83,15 @@ export type SignedNewAttestation = (
     block: string,
     index: number,
     slot: number,
-    bits: string,
+    bitsIndex: number,
 ) => {
-    payload: {block: string; index: number; slot: number; bits: string};
+    payload: {block: string; index: number; slot: number; bitsIndex: number};
     meta: string;
 };
 export const signedNewAttestation = createAction<SignedNewAttestation>(
     "validator/signedNewAttestation",
-    (publicKey: string, block: string, index: number, slot: number, bits: string) => ({
-        payload: {block, index, slot, bits},
+    (publicKey: string, block: string, index: number, slot: number, bitsIndex: number) => ({
+        payload: {block, index, slot, bitsIndex},
         meta: publicKey,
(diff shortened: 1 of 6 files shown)


src/renderer/ducks/validator/actions.ts
```diff
@@ -83,14 +83,15 @@ export type SignedNewAttestation = (
...
-    payload: {block: string; index: number; slot: number};
+    payload: {block: string; index: number; slot: number; bits: string};
     meta: string;
 };
 export const signedNewAttestation = createAction<SignedNewAttestation>(
     "validator/signedNewAttestation",
-    (publicKey: string, block: string, index: number, slot: number) => ({
-        payload: {block, index, slot},
+    (publicKey: string, block: string, index: number, slot: number, bits: string) => ({
+        payload: {block, index, slot, bits},
```
mpetrunic, line 94: you don't need bits here. YOu just need index of true bit in attestation (new attestation must have only 1 bit set - on index matching validator position in committee).

src/renderer/ducks/validator/sagas.ts
```diff
@@ -445,17 +446,19 @@ export function* getAttestationEffectiveness({
                 skippedQue++;
             } else {
                 const sanitizedAttestations = result.filter(
-                    ({data}) => payload.block === toHex(data.beaconBlockRoot) && data.index === payload.index,
+                    ({data, aggregationBits}) =>
+                        payload.block === toHex(data.beaconBlockRoot) &&
+                        data.index === payload.index &&
+                        data.slot === payload.slot &&
+                        payload.bits === JSON.stringify(aggregationBits),
```
mpetrunic, line 453: You need to check if bit on validator position in committee is set to true. I think you should also check source and target root.

mpetrunic (changes requested):

src/renderer/ducks/validator/sagas.ts
BeroBurny: `source` and `target` contains `epoch` and `root` of `slot` as i check `source` `slot` in parent object of `source` and think checking for `source` and `target` can be bit a overkill and add unnecessary complexity

mpetrunic: it contains epoch and epoch boundary root (it's basically validator vote) that are unreleated to slot. That said since, you are checking committee index, it's highly unlikely that someone created attestation with different source and target outside of chainguardian (in that case more pressing issue is slashing anyways).

src/renderer/ducks/validator/actions.ts
```diff
@@ -83,14 +83,15 @@ export type SignedNewAttestation = (
     block: string,
     index: number,
     slot: number,
+    bitsIndex: number,
 ) => {
-    payload: {block: string; index: number; slot: number};
+    payload: {block: string; index: number; slot: number; bitsIndex: number};
```
mpetrunic, line 88:
```suggestion
    payload: {block: string; index: number; slot: number; validatorIndexInCommittee: number};
```

src/renderer/ducks/validator/sagas.ts
```diff
@@ -445,30 +445,31 @@ export function* getAttestationEffectiveness({
                 skippedQue++;
             } else {
                 const sanitizedAttestations = result.filter(
-                    ({data}) => payload.block === toHex(data.beaconBlockRoot) && data.index === payload.index,
+                    ({data, aggregationBits}) =>
+                        payload.block === toHex(data.beaconBlockRoot) &&
+                        data.index === payload.index &&
+                        data.slot === payload.slot &&
+                        payload.bitsIndex === aggregationBits.findIndex((bit) => bit),
```
mpetrunic, line 452:
```suggestion
                        aggregationBits[validatorIndexInCommittee] === true
```

src/renderer/services/eth2/client/eth2ApiClient/cgEth2BeaconPoolApi.ts
```diff
@@ -24,6 +24,7 @@ export class CgEth2BeaconPoolApi implements IBeaconPoolApi {
                     toHex(attestation.data.beaconBlockRoot),
                     attestation.data.index,
                     attestation.data.slot,
+                    attestation.aggregationBits.findIndex((bit) => Number(bit) === -1),
```
mpetrunic, line 27:
```suggestion
                    attestation.aggregationBits.findIndex((bit) => bit === true),
```

src/renderer/ducks/validator/actions.ts
mpetrunic: you should maybe throw error if this is < 0 (Don't think it can happen but to be on the safe side)

BeroBurny: how we should handle this error if happens?

src/renderer/services/eth2/client/eth2ApiClient/cgEth2BeaconPoolApi.ts
BeroBurny: we talk about this... for some reason we got `-true` what is `-1`, if i try to compare `bit === -true` got type error i cannot compare boolean with number

src/renderer/ducks/validator/actions.ts
BeroBurny: i can add case if is index `-1` (not found) to not call saga

src/renderer/services/eth2/client/eth2ApiClient/cgEth2BeaconPoolApi.ts
mpetrunic: I don't think that's possible could you log aggregation bits and post logs here?

mpetrunic (approved):

Issue #529, opened 2021-03-09
Support Teku BN


morrigan:

BeroBurny:
for implementation use this guidelines https://github.com/NodeFactoryIo/ChainGuardian/pull/541#discussion_r597486236

PR #541, opened 2021-03-18
Localnet validator testing

### LINKED ISSUE #528: Setup validator testing with local network
Further instructions will be given by @mpetrunic

BeroBurny:
**Short description of work done**

- implement validator tester on beacon node

### PR Checklist

- [ ] I have run type check locally
- [x] I have run linter locally
- [ ] I have run unit and integration tests locally
- [x] Rebased to master branch / merged master

### Changes

- implement validator tester on beacon node
- update lighthouse files
- implement event source polyfill (needed for test)

### Issues

Closes #528


integration/lighthouse.ts
```diff
@@ -0,0 +1,20 @@
+import {SecretKey} from "@chainsafe/bls";
+import {Keystore} from "@chainsafe/bls-keystore";
+import keystore from "./lighthouse-keystore.json";
+import assert from "assert";
+
+import {restValidation} from "./restValidation";
+const sk = "222222222222222222222222222222222222222222222222222";
```
mpetrunic, line 7: password?

integration/restValidation.ts
```diff
@@ -0,0 +1,215 @@
...
+import {Validator} from "@chainsafe/lodestar-validator";
+import {config} from "@chainsafe/lodestar-config/lib/presets/mainnet";
+import {CgEth2ApiClient} from "../src/renderer/services/eth2/client/eth2ApiClient";
+import {CGSlashingProtection} from "../src/renderer/services/eth2/client/slashingProtection";
+import sinon from "sinon";
+import EventSource from "eventsource";
+import {computeEpochAtSlot} from "@chainsafe/lodestar-beacon-state-transition";
+import axios from "axios";
+
+const getValidatorIndexFromBits = (aggregationBits: string): number => {
```
mpetrunic, line 12:
```suggestion
const getValidatorIndexInCommitteeFromBits = (aggregationBits: string): number => {
```

integration/restValidation.ts
```diff
@@ -0,0 +1,215 @@
...
+    index: number;
+    slot: number;
+    validators: number[];
+};
+const getCommitteesFactory = (baseUrl: string) => async (
+    validatorIndex: number,
+    blockSlot: number,
+    ignoreBefore?: number,
+): Promise<Committees[]> => {
+    const url = new URL(`/eth/v1/beacon/states/${blockSlot}/committees`, baseUrl);
```
mpetrunic, line 40: can we use beacon api client rather than re-implementing everywhere?

API will receive changes over time and it will be nightmare to update on bunch of places.

integration/restValidation.ts
```diff
@@ -0,0 +1,215 @@
...
+        delegated: number;
+    };
+    attestation: {
+        attestations: number;
+        delegated: number;
+    };
+}> =>
+    new Promise((resolve) => {
+        (async (): Promise<void> => {
+            await initBLS("herumi");
```
mpetrunic, line 115:
```suggestion
            await initBLS("blst-native");
```
is faster

integration/restValidation.ts
```diff
@@ -0,0 +1,215 @@
...
+            const getProposer = getProposerFactory(baseUrl);
+            const isBlockProposed = isBlockProposedFactory(baseUrl);
+            const proposers = new Map<number, boolean>();
+
+            const validatorUrl = new URL(`/eth/v1/beacon/states/head/validators/${validatorPublicKey}`, baseUrl);
+            const validator = await axios.get(validatorUrl.href);
+            const validatorIndex = Number(validator.data.data.index);
+
+            const url = new URL(`/eth/v1/events?topics=attestation,head`, baseUrl);
+            const eventSource = new EventSource(url.href);
```
mpetrunic, line 152: you already have eth2API instance why not use that to listen on events?

BeroBurny: probably is this problem? https://github.com/ChainSafe/ssz/issues/23
I tried validator on `Teku` Beacon node and got some error while `szz` deserializing

integration/lighthouse.ts
BeroBurny: its imported validator Keystore generated by lighthouse which contains secret key like this

mpetrunic: from what I see in code you are decrypting secret key from keystore and use this variable as password for decryption

integration/restValidation.ts
mpetrunic: I think we should extend CgEth2Api with specific implementation for different clients until they all converge to be fully compliant with spec.

Like `CgTekuEth2Api extends CgEth2Api` and `CgLighthouseEth2Api extends CgEth2Api` where CgEth2Api is spec compliant implementation

mpetrunic: > probably is this problem? ChainSafe/ssz#23
I tried validator on Teku Beacon node and got some error while szz deserializing

Nope, you shouldn't get empty beacon blocks over API. Send me json you get from Teku, there is probably some discrepancies with data we get.

integration/restValidation.ts
BeroBurny: what if the validator beacon API master class has something wrongly implemented?

mpetrunic: we override if there is non compliant api: https://github.com/NodeFactoryIo/ChainGuardian/pull/541#discussion_r597486236

integration/lighthouse.ts
BeroBurny: `sk` means secret key

mpetrunic: My point is that variable name should be password haha

integration/restValidation.ts
BeroBurny: > I think we should extend CgEth2Api with specific implementation for different clients until they all converge to be fully compliant with spec.
> 
> Like `CgTekuEth2Api extends CgEth2Api` and `CgLighthouseEth2Api extends CgEth2Api` where CgEth2Api is spec compliant implementation

and this will cause the problem mentioned above https://github.com/NodeFactoryIo/ChainGuardian/pull/541#discussion_r597473300
then will be better to provide `client` and `version` to the client and if there is some non-compliant we can override for specific (or range) `client`/`version`

mpetrunic: But you know what client you will use for each testnet so you can provide that info in this PR and instantiate appropriate beacon api client instance. As for chainguardian, ideas was (and should still be) to query https://ethereum.github.io/eth2.0-APIs/#/Node/getNodeVersion and depending on that initialize either LighthouseBeaconApi instance or TekuBeaconApi instance or throw error if not supported.

integration/restValidation.ts
BeroBurny: wont fix for some reason i got
> Require stack:
> - /home/bernard/WebstormProjects/ChainGuardian/node_modules/@chainsafe/bls/lib/blst/secretKey.js
> - /home/bernard/WebstormProjects/ChainGuardian/node_modules/@chainsafe/bls/lib/blst/index.js
> - /home/bernard/WebstormProjects/ChainGuardian/node_modules/@chainsafe/bls/lib/index.js

integration/restValidation.ts
BeroBurny: it is `out of scope` for this task
for Teku there is #529 where we will proceed by this https://github.com/NodeFactoryIo/ChainGuardian/pull/541#discussion_r597486236

integration/restValidation.ts
mpetrunic: thats stack, but what's the error? 😄

integration/restValidation.ts
mpetrunic: It makes no sense to merge this PR with all those new beacon api implementation. If this Pr is only for Lighthouse testing, than just use CgEth2BeaconApi everywhere and we can have separate PR with teku specific implementation.

BeroBurny: then what we gonna do with #529  ?

integration/restValidation.ts
BeroBurny:
```sh
yarn run v1.22.5
$ node -r ts-node/register integration/lighthouse.ts
(node:3219949) UnhandledPromiseRejectionWarning: Error: Cannot find module '@chainsafe/blst'
Require stack:
- /home/bernard/WebstormProjects/ChainGuardian/node_modules/@chainsafe/bls/lib/blst/secretKey.js
- /home/bernard/WebstormProjects/ChainGuardian/node_modules/@chainsafe/bls/lib/blst/index.js
- /home/bernard/WebstormProjects/ChainGuardian/node_modules/@chainsafe/bls/lib/index.js
- /home/bernard/WebstormProjects/ChainGuardian/integration/lighthouse.ts
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:815:15)
    at Function.Module._load (internal/modules/cjs/loader.js:667:27)
    at Module.require (internal/modules/cjs/loader.js:887:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.<anonymous> (/home/bernard/WebstormProjects/ChainGuardian/node_modules/@chainsafe/bls/lib/blst/secretKey.js:10:27)
    at Module._compile (internal/modules/cjs/loader.js:999:30)
    at Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
    at Object.require.extensions.<computed> [as .js] (/home/bernard/WebstormProjects/ChainGuardian/node_modules/ts-node/src/index.ts:851:44)
    at Module.load (internal/modules/cjs/loader.js:863:32)
    at Function.Module._load (internal/modules/cjs/loader.js:708:14)
(node:3219949) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 2)
(node:3219949) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
Done in 4.70s.
```

mpetrunic: ahh, yarn add @chainsafe/blst it's optional dependency. 

Not sure if that works with electron when compiled into binary.
Let's use herumi and maybe open issue to experiment adding blst.

integration/restValidation.ts
mpetrunic: Ok then, just use CgEth2BeaconApi here (in this PR) and add other api client implementations as part of #529 

Log spec discrepancies so we can report it upstream.

integration/restValidation.ts
BeroBurny: think it is problem in `ts-node/register` like i needed to add `eventsource` polyfill to make this tester work (and some other stuff)

mpetrunic: For error you posted `yarn add @chainsafe/blst` should work.

As for EventSource not sure if you even need it if you use CgEth2BeaconApi

BeroBurny: its used under `CgEth2EventsApi` :man_shrugging:

mpetrunic: how did it work before then? 😄

BeroBurny: probably magic :man_shrugging: JavaScript magic :mage:

BeroBurny: > For error you posted `yarn add @chainsafe/blst` should work.
> 
> As for EventSource not sure if you even need it if you use CgEth2BeaconApi

and works

integration/lighthouse.ts
```diff
@@ -0,0 +1,20 @@
+import {SecretKey} from "@chainsafe/bls";
+import {Keystore} from "@chainsafe/bls-keystore";
+import keystore from "./lighthouse-keystore.json";
+import assert from "assert";
+
+import {restValidation} from "./restValidation";
+const secretKey = "222222222222222222222222222222222222222222222222222";
```
mpetrunic, line 7:
```suggestion
const keystorePassword = "222222222222222222222222222222222222222222222222222";
```

integration/restValidation.ts
```diff
@@ -0,0 +1,162 @@
...
+            const getProposer = getProposerFactory(eth2API);
+            const isBlockProposed = isBlockProposedFactory(eth2API);
+            const proposers = new Map<number, boolean>();
+
+            const stream = await eth2API.events.getEventStream(([
+                CGBeaconEventType.HEAD,
+                CGBeaconEventType.ATTESTATION,
+            ] as unknown) as BeaconEventType[]);
+            for await (const {type, message} of stream) {
+                if (((type as unknown) as CGBeaconEventType.ATTESTATION) === CGBeaconEventType.ATTESTATION) {
```
mpetrunic, line 107: to make it more readable, could you move each case implementation to separate function?

src/renderer/services/api/http/httpClient.ts
```diff
@@ -59,20 +59,22 @@ export class HttpClient {
         const start = Date.now();
         return {
             onComplete: ({request, status, config}: AxiosResponse): void => {
-                database.networkMetrics.addRecord(config.baseURL, {
-                    url: request.responseURL,
-                    code: status,
-                    latency: Date.now() - start,
-                    time: Date.now(),
-                });
+                if (process.env.NODE_ENV)
```
mpetrunic, line 62: == "production" ?

mpetrunic (commented):
LGTM, few more details

src/renderer/services/api/http/httpClient.ts
BeroBurny: this is intentional
we want this to be active in `production` and `development` but not when is used validator testing, when you start validator testing you use `ts-node` which does not provide this `NODE_ENV` variable

integration/restValidation.ts
BeroBurny: it's hard to do that, as this cases depends on the upper scope

src/renderer/services/api/http/httpClient.ts
mpetrunic: It doesn't by default but it will send value if somebody has that env exported for some reason. I would be explicit here.

integration/restValidation.ts
mpetrunic: Can't you pass required params as method args?

BeroBurny: `CGBeaconEventType.ATTESTATION` just use Map `attestations`
but `CGBeaconEventType.HEAD` uses from uper scope
- slot
- proposers
- firstSlot
- startEpoch
- lastEpoch
- committees
- proposers
- attestations  

pretty needy case :see_no_evil:

mpetrunic: pass everything as huge object. It's really hard to follow that huge piece of code 😅

BeroBurny: and then return as the very huge object to assign all variables to upper scope variables?

mpetrunic (approved):

